PR #17, opened 2017-09-11
Prep for 1.1.1 release with net40 support.

mikeng13:
@optimizely/fullstack-devs


mikeng13:
Yep, as discussed offline, will check it out and circle back in another diff.

PR #19, opened 2017-09-27
Kroach/oasis 1806 null

kellyroach-optimizely:
OASIS-1806 [C#] Prune null/nil/None value attributes in track/activate

Don't include JSON null's in any JSON the SDK sends to server.

Among the reasons why is the distinction between actual absence
of a key-value pair and presence of a key-value pair whose value
is "null" gets messy/confused/unsupported depending on the particular
platform or DB . Limit Optimizely JSON's to JSON's built via
booleans, numbers, strings, arrays, and objects (aka dictionaries).
Don't use JSON null's.

### DIFF
--- a/OptimizelySDK.DemoApp/OptimizelySDK.DemoApp.csproj
+++ b/OptimizelySDK.DemoApp/OptimizelySDK.DemoApp.csproj
@@ -28,8 +28,8 @@
     <TargetFrameworkProfile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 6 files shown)


kellyroach-optimizely:
TEST:

"Unit Tests" / "Run All"
Passed: 92 Failed: 0
including our 2 new tests:
TestActivateWithNullAttributes
TestTrackWithNullAttributesWithNullEventValue

kellyroach-optimizely:
TEST:
* "Unit Tests" / "Run All"
Passed: 92 Failed: 0

mikeng13:
@kellyroach-optimizely go ahead and Squash Merge this. The failing e2e tests are the event tag tests that we are already looking into and aren't related to this diff. The rest of the tests pass.

PR #24, opened 2017-10-12
Folio3/bucketing

alda-optimizely:
Pulling the bucketing id PR from Folio 3 locally so that I can run e2e tests.


mikeng13:
build

PR #32, opened 2017-11-09
snk now only mentioned by pack.ps1

kellyroach-optimizely:
This P.R. leaves task of calling sn.exe and strongnaming entirely in pack.ps1's hands.

There were some new issues with the redundant and not wholely effective calls to sn.exe
in VS 2017 custom "After Build" commands.  One issue is these cause "Build FAILED"s
in our E2E tests which rely on '/bin/sh -c dotnet build --configuration Release'
instead of msbuild.

The E2E regression goes away in this P.R.

### DIFF
--- a/OptimizelySDK.Net35/OptimizelySDK.Net35.csproj
+++ b/OptimizelySDK.Net35/OptimizelySDK.Net35.csproj
@@ -12,8 +12,6 @@
     <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <ReleaseVersion>1.2.1</ReleaseVersion>
-    <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>..\keypair.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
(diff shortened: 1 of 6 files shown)


OptimizelySDK.Net35/OptimizelySDK.Net35.csproj
```diff
@@ -12,8 +12,6 @@
     <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <ReleaseVersion>1.2.1</ReleaseVersion>
-    <SignAssembly>true</SignAssembly>
```
wangjoshuah: are we removing signing?

kellyroach-optimizely: No, we are still signing.  The pack.ps1 does the sn.exe signing.  There seem to be about 4 different ways to do sn.exe signing documented at various MS URL's, but the most reliable approach for us has been sn.exe in pack.ps1 .  We're keeping this.  The other 2 ways we're dropping were redundancies which, unfortunately, do interfere with our E2E testing.

PR #49, opened 2018-01-04
Kroach/oasis 2163 logger

kellyroach-optimizely:
OASIS-2163 [C#] 1.3.0 Release Candidate Patch DefaultLogger Issue

C# 1.3.0 DefaultLogger wasn't logging anything.  FIXED.

THREE THINGS WORTH MENTIONING:
* DefaultLogger (DefaultLogger.cs, ILogger.cs, Optimizely.cs)
The DefaultLogger is distributed in Release *.dll in a *.nupkg, so can't use Debug.WriteLine .
It is switched now to using Console.WriteLine which addresses OASIS-2163 issue.
Since the DefaultLogger is now more talkative, we did add a Verbosity control
on how much it talks.  The Optimizely's constructor default logger is now the
DefaultLogger, not the NoOpLogger .
* DemoApp (README.md, Home/Index.cshtml)
There are improvements in the DemoApp's README.md and Home/Index.cshtml
wrt appearance and information.  The Home/Index.cshtml "Learn more" buttons now
work.  DemoApp changes don't go into the *.nupkg but are available from GITHUB.COM .
* 4 *.csproj sn.exe signing changes.  Our E2E doesn't understand these.  Going forward,
we will turn on the 4 checkboxes on our local machine when we build *.nupkg release
candidates.  Our QA process already is checking for strong naming and digital signing
of our *.nupkg *.dll's and this won't change.

This branch will merge onto the "1.3.x" branch which is code that will be compiled
to produce a new "1.3.0" *.nupkg we plan to release.


kellyroach-optimizely:
TEST:
* Clean All / Build All PASSED
* 140/140 unit tests PASSED
* E2E customized for 1.3.0 PASSED
* DemoApp PASSED

OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
```diff
@@ -3,7 +3,6 @@
     <PropertyGroup>
         <TargetFramework>netstandard1.6</TargetFramework>
         <ReleaseVersion>1.2.1</ReleaseVersion>
-        <SignAssembly>true</SignAssembly>
```
wangjoshuah: are we no longer signing the distributions?

OptimizelySDK/Logger/DefaultLogger.cs
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-using System.Diagnostics;
+using System;
```
wangjoshuah: where is `System` used?

OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
kellyroach-optimizely: We're still signing the distributions, but the process we follow is changed.
To sn.exe (typically we only do this during prep of a *.nupkg), we either check on 4 Options checkboxes in VS on our local Mac machine or run a *.ps1 script.  The strong naming and digital signing we put in place for our *.nupkg releases remains and is still QA checked that it happens.

kellyroach-optimizely: The change is necessitated by E2E which would otherwise generate "CSC : error CS7027: Error signing output with public key from file" errors.  If it sounds a bit familiar, it is because it an issue we've been wrestling with and perfecting.

OptimizelySDK/Logger/DefaultLogger.cs
kellyroach-optimizely: "Console" class is defined in the "System" namespace.

kellyroach-optimizely: Line 38 of DefaultLogger.cs is "Console.WriteLine(line);"

wangjoshuah: Are we linking/importing all system files here like in Java? Would System.Console be better?

kellyroach-optimizely:
NOTE: red cross on "default -- Build finished" came from the "master" branch fullstack-e2e tests which are not 100% in synch with the "1.3.0" API's.  We know this and it is not a concern to us.  The fullstack-e2e modified for "1.3.0" API's PASSED.

J.W. APPROVED.  Merging onto 1.3.x branch.

Issue #51, opened 2018-01-08
Timing issue with Activate() & Track() being called sequentially while not able to await on the async method calls


HistoricallyCorrect:
I hit a bug with Optimizely C# SDK when toying in "short lived" console app. The issue is mainly that the SDK does not employ C# async/await best practice "**Async All the Way**". Reference can be [here](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx).

The problematic code is [here](https://github.com/optimizely/csharp-sdk/blob/dc40c801fef3cb9d68d1a80179f06caf3c97d09e/OptimizelySDK/Event/Dispatcher/HttpClientEventDispatcher45.cs#L77) and the usages are [here](https://github.com/optimizely/csharp-sdk/search?utf8=%E2%9C%93&q=EventDispatcher.DispatchEvent&type=). And the two scenarios that I came up with are as below:

```C#
//Scenario 1
var client = new Optimizely(dataFile);
client.Activate("experimentId", "userId", attrs);
//Activate() will call SendImpressionEvent() which calls EventDispatcher.DispatchEvent(impressionEvent); But Activate() does not have an async method signature, hence not possible to await on the response.

client.Track("Clicks", "userId");
//By the time Track() is called, if above DispatchEvent is still in-flight (not complete), I'm assuming Track() will fail due to "userId" has not been activated for "experimentId".

```

```C#
//Scenario 2
var client = new Optimizely(dataFile);
client.Activate("experimentId", "userId", attrs);
client.Track("Clicks", "userId");
//At this point in time, my console app exits, because Track() does not have an async method signature, hence I'm not able to tell if the DispatchEvent inside Track() has completed successfully ot not.
```

I'll also submit a PR for fix soon.

**Update**:

Having had a bit more time to think about this issue, I believe it is best for DispatchEvent() to be sequential and not fire & forget. For fire & forget, it is best to be more explicit in terms of method signature.

Proposed change is here:

https://github.com/HistoricallyCorrect/csharp-sdk/commit/44ffbbfd2598bcd9a60d678b88d1e667d1865eab

HistoricallyCorrect:
@mikeng13

kellyroach-optimizely:
Thank you for submitting your issue, the care with which you wrote it, and the suggested code change which you believe would resolve the issue.  We will take a careful look at what you've written.  I've included all the information you've supplied in an internal Optimizely JIRA ticket which we will study and follow up on.

[...]

PR #52, opened 2018-01-09
DispatchEvent() must be sequential and cannot be fire & forget

HistoricallyCorrect:
When Activate() & Track() are called, Track() can only proceed after
Activate() completes successfully. Hence the fire & forget pattern for
DispatchEvent() will not work.

Original issue:

https://github.com/optimizely/csharp-sdk/issues/51

### DIFF
--- a/OptimizelySDK/Event/Dispatcher/HttpClientEventDispatcher45.cs
+++ b/OptimizelySDK/Event/Dispatcher/HttpClientEventDispatcher45.cs
@@ -68,16 +68,6 @@ namespace OptimizelySDK.Event.Dispatcher
         }
 
         /// <summary>
-        /// Dispatch an event Asynchronously by creating a new task and calls the 
-        /// Async version of DispatchEvent
-        /// This is a "Fire and Forget" option
-        /// </summary>
-        public void DispatchEventSilently(LogEvent logEvent)
-        {
-            Task.Run(() => DispatchEventAsync(logEvent));
-        }
-
-        /// <summary>
         /// Dispatch an event Asynchronously by creating a new task and calls the
         /// Async version of DispatchEvent
         /// This is a "sequential" option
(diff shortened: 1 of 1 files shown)


HistoricallyCorrect:
@mikeng13

OptimizelySDK/Event/Dispatcher/HttpClientEventDispatcher45.cs
```diff
@@ -72,10 +72,20 @@ static HttpClientEventDispatcher45()
         /// Async version of DispatchEvent
         /// This is a "Fire and Forget" option
         /// </summary>
-        public void DispatchEvent(LogEvent logEvent)
+        public void DispatchEventSilently(LogEvent logEvent)
         {
             Task.Run(() => DispatchEventAsync(logEvent));
         }
+
```
kellyroach-optimizely: The prime effect of the P.R. #52 proposal will be to change the body of DispatchEvent .
However, I don't understand the purpose of keeping DispatchEventSilently around.
The SDK code doesn't call it and HttpClientEventDispatcher45 isn't documented in
Optimizely's developer documentation.  Isn't DispatchEventSilently dead code?  Why
isn't this deleted?

HistoricallyCorrect: Because I was not sure whether there was a legit reason behind DispatchEvent hiding asynchronous nature of DispatchEventAsync. So I left and renamed the old code as DispatchEventSilently thinking maybe you want to preserve a async version of DispatchEvent and name it more obviously. But if there is no strong reason behind it, then I agree DispatchEventSilently is dead code and should be removed.

kellyroach-optimizely:
SUPERCEDED BY APPROVED P.R. #54

PR #70, opened 2018-04-03
use case insensitive comparer to sort the return array

wangjoshuah:


mikeng13:
This is no longer needed. Going to close out.

PR #80, opened 2018-06-06
Enable source clear

aliabbasrizvi:


mikeng13:
Hmm the build failure on Travis seems like a legit blocker: `The command "curl -sSL https://download.sourceclear.com/ci.sh |  bash" failed and exited with 1 during .`

aliabbasrizvi:
Yeah, looking into it.

PR #93, opened 2018-08-09
Adds IOptimizely interface

### LINKED ISSUE #92: Give the Optimizely client an interface to facilitate testing
It is very difficult to write unit tests around our code that makes use of Optimizely, since it provides no interface from which to create a suitable mock. We can make do by creating our own wrapping class, but this gets messy quickly.

Can an `IOptimizely` interface be added, so that we can use NSubstitute or a similar library to mock the Optimizely client?

Nevett:
Fixes #92.

Extracted an interface for all public members.

Also fixed a couple of the xmldocs.

### DIFF
--- a/OptimizelySDK/IOptimizely.cs
+++ b/OptimizelySDK/IOptimizely.cs
@@ -1,0 +1,140 @@
+﻿/**
+ * Copyright 2018, Optimizely, Inc. and contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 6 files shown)


Nevett:
Done!

OptimizelySDK/IOptimizely.cs
```diff
@@ -0,0 +1,140 @@
...
+		/// <summary>
+		/// Gets the feature variable value for given type.
+		/// </summary>
+		/// <param name="featureKey">The feature flag key</param>
+		/// <param name="variableKey">The variable key</param>
+		/// <param name="userId">The user ID</param>
+		/// <param name="userAttributes">The user's attributes</param>
+		/// <param name="variableType">Variable type</param>
+		/// <returns>string | null Feature variable value</returns>
+		string GetFeatureVariableValueForType(string featureKey, string variableKey, string userId,
```
msohailhussain: Please remove this method, it is being used internally. no interface needed.

OptimizelySDK/Optimizely.cs
```diff
@@ -360,7 +360,7 @@ public Variation GetForcedVariation(string experimentKey, string userId)
         /// Determine whether a feature is enabled.
         /// Send an impression event if the user is bucketed into an experiment using the feature.
         /// </summary>
-        /// <param name="experimentKey">The experiment key</param>
+        /// <param name="featureKey">The experiment key</param>
```
msohailhussain: Please change 'experiment' to 'feature' in description.

PR #98, opened 2018-08-21
Whitelisted bucketing logic is in DecisionService

msohailhussain:
We are checking whitelisted users in decision service, get variation method. This logic was placed when DecisionService class didn't exist.


mikeng13:
build

mikeng13:
build-e2e

PR #100, opened 2018-08-22
fix(track): Send decisions for all experiments using an event when using track

msohailhussain:
fix(track): Send decisions for all experiments using an event when using track.


mikeng13:
build

OptimizelySDK.Tests/MultiEventCachedData.json
```diff
@@ -0,0 +1,186 @@
+﻿{
```
mikeng13: This name is very specific to this scenario. Instead, I can see us potentially using this same datafile for more sophisticated tests. Let's call the datafile `simple_ab_experiments` (later we can have one called `feature_experiments`.

In the datafile itself, we should call the event --> `event_with_multiple_running_experiments`

The experiment themselves can be named more descriptively like:
- `running_experiment_1`
- `running_experiment_2`

This makes it easier for us to reuse some of the entities in this test datafile

mfahadahmed: Done.

PR #101, opened 2018-08-29
fix(datafile-parsing): Prevent newer versions datafile

msohailhussain:


OptimizelySDK/Optimizely.cs
```diff
@@ -111,9 +112,15 @@ public class Optimizely : IOptimizely
                 IsValid = true;
                 DecisionService = new DecisionService(Bucketer, ErrorHandler, Config, userProfileService, Logger);
             }
+            catch (ConfigParseException configException)
+            {
+                Logger.Log(LogLevel.ERROR, configException.Message);
```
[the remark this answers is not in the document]
mfahadahmed: Done.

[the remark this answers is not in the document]
mikeng13: Let's consolidate these two catches into one and handle the logging based on the type of the exception. We can avoid duplicating some of the code, like passing the error to the error handler

OptimizelySDK/ProjectConfig.cs
```diff
@@ -280,6 +296,24 @@ public static ProjectConfig Create(string content, ILogger logger, IErrorHandler
...
+        {
+            if (configData == null)
+                throw new ConfigParseException("Unable to parse null datafile.");
+
+            if (string.IsNullOrEmpty(configData))
+                throw new ConfigParseException("Unable to parse empty datafile.");
+
+            var config = JsonConvert.DeserializeObject<ProjectConfig>(configData);
+
+            if (SupportedVersions.TrueForAll((obj) => !(((int)obj).ToString() == config.Version)))
```
mikeng13: Please don't use the generic variable name `obj`. Instead go for `supportVersion`

OptimizelySDK.Tests/ProjectConfigTest.cs
```diff
@@ -863,5 +863,36 @@ public void TestGetAttributeIdWithInvalidAttributeKey()
...
+        public void TestCreateThrowsWithEmptyDatafile()
+        {
+            var exception = Assert.Throws<ConfigParseException>(() => ProjectConfig.Create("", null, null));
+            Assert.AreEqual("Unable to parse empty datafile.", exception.Message);
+        }
+
+        [Test]
+        public void TestCreateThrowsWithUnsupportedDatafileVersion()
+        {
+            var unsupportedConfig = JsonConvert.DeserializeObject<ProjectConfig>(TestData.Datafile);
```
mikeng13: Please create a separate mock unsupported datafile and load it into this test case

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -185,7 +186,23 @@ public void TestValidateInputsInvalidFileJsonValidationSkipped()
         {
             string datafile = "{\"name\":\"optimizely\"}";
             Optimizely optimizely = new Optimizely(datafile, null, null, null, skipJsonValidation: true);
-            Assert.IsTrue(optimizely.IsValid);
+            Assert.IsFalse(optimizely.IsValid);
+        }
+
+        [Test]
+        public void TestErrorHandlingWithInvalidConfigVersion()
```
mikeng13: Please split these up into three separate unit tests. Makes it easier to read and determine what actually failed

PR #102, opened 2018-09-12
feat(api): Accepting all types for attributes values

mfahadahmed:


OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -458,6 +458,39 @@ public void TestActivateExperimentNotRunning()
 
             Assert.IsNull(variationkey);
         }
+
+        [Test]
+        public void TestActivateWithTypedAttributes()
```
msohailhussain: This test case is not verifying the payload generated from EventBuilder is returned correctly or not. 
Create an expectedEventBuilder class and call EventBuilder, and compare it.

OptimizelySDK/Bucketing/DecisionService.cs
```diff
@@ -393,16 +393,23 @@ public virtual FeatureDecision GetVariationForFeature(FeatureFlag featureFlag, s
...
         private string GetBucketingId(string userId, UserAttributes filteredAttributes)
         {
             string bucketingId = userId;
 
             // If the bucketing ID key is defined in attributes, then use that in place of the userID for the murmur hash key
             if (filteredAttributes != null && filteredAttributes.ContainsKey(ControlAttributes.BUCKETING_ID_ATTRIBUTE))
             {
-                bucketingId = filteredAttributes[ControlAttributes.BUCKETING_ID_ATTRIBUTE];
-                Logger.Log(LogLevel.DEBUG, string.Format("Setting the bucketing ID to \"{0}\"", bucketingId));
+                if (filteredAttributes[ControlAttributes.BUCKETING_ID_ATTRIBUTE] is string)
```
msohailhussain: Please combine these two lines. 

(filteredAttributes[ControlAttributes.BUCKETING_ID_ATTRIBUTE] as string ?? userId).ToString()
should serve the purpose, please check.

OptimizelySDK/Utils/ConditionEvaluator.cs
```diff
@@ -71,7 +71,7 @@ public bool Evaluate(JToken conditions, UserAttributes userAttributes)
 
             string conditionName = conditions["name"].ToString();
             return userAttributes != null && userAttributes.ContainsKey(conditionName) 
-                && userAttributes[conditionName] == conditions["value"].ToString();
+                && userAttributes[conditionName].ToString() == conditions["value"].ToString();
```
msohailhussain: This may cause some problem when any value other than string is converted into string. @oakbani

OptimizelySDK/Bucketing/DecisionService.cs
msohailhussain: leave it, as is used for casting not for type checking.

OptimizelySDK.Tests/OptimizelyTest.cs
mfahadahmed: Added payload comparison test case for typed attributes in EventBuilderTest.cs.

Issue #78, opened 2018-06-01
JsonNet.PrivateSettersContractResolvers.Source dependency is leaky. 


jwellman-uship:
Consider pulling this source directly into the project or relying on a dll instead if one is available.  Since the source is being pulled in that will pull this source into any project that relies on the Optimizely nuget anywhere down stream.

artempushkin:
+1. The source file conflicts with our code validation rules.

mikeng13:
Thanks for reporting. This is now fixed in https://github.com/optimizely/csharp-sdk/releases/tag/v2.2.1

PR #105, opened 2018-10-19
Remove JsonNet.PrivateSettersContractResolvers dependency.

mfahadahmed:
we removed this dependency as it is not longer being used in Optimizely Project. It was also added as developmentDependency in nuspec which is not really needed. 
Removed developmentdepency from demoapp and from OptimizelySDK package.config file and also from nuspec.


msohailhussain:
https://github.com/optimizely/csharp-sdk/issues/78

PR #112, opened 2018-11-08
feat(eventdispatcher): HttpClientEventDispatcher45 now logs full exception

mfahadahmed:


OptimizelySDK/Utils/ExtensionMethods.cs
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.Text;
+
+namespace OptimizelySDK.Utils
+{
+    public static class ExtensionMethods
```
mikeng13: This seems like a very generic name. Can we call it `ExceptionExtensionMethods`?

This is actually a pretty cool feature of the platform/language btw.

msohailhussain: @mikeng13 I believe the name should be ExceptionDetailExtension, since it's class, looks odd to call with Methods? make sense?

mfahadahmed: Renamed the class as ExceptionExtensions to contain any exception related extensions in it. Thoughts?

mfahadahmed: Also making the class internal restricts us from writing its unit tests.

OptimizelySDK/Utils/ExceptionExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace OptimizelySDK.Utils
+{
+    public static class ExceptionExtensions
```
msohailhussain: internal?

msohailhussain: got it.

OptimizelySDK.Tests/UtilsTests/ExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,34 @@
+﻿/* 
+ * Copyright 2018, Optimizely
```
mikeng13: 2019 now :)

OptimizelySDK/Event/Dispatcher/HttpClientEventDispatcher45.cs
```diff
@@ -1,5 +1,5 @@
 ﻿/* 
- * Copyright 2017, Optimizely
+ * Copyright 2017-2018, Optimizely
```
mikeng13: 2019 :)

PR #117, opened 2018-12-20
feat (audience match types): Update condition evaluator for new audience match types.

mfahadahmed:
### Summary

This adds support for new audience match type conditions to the condition evaluator:
- exact, exists, gt, lt, and substring conditions
- Abort and return null when appropriate in leaf evaluators
- Null handling in and/or/not evaluators

### Test Plan

Added new unit tests.

### DIFF
--- a/OptimizelySDK/Utils/Evaluator.cs
+++ b/OptimizelySDK/Utils/Evaluator.cs
@@ -1,0 +1,102 @@
+﻿/* 
+ * Copyright 2018, Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


msohailhussain:
@mfahadahmed we have to follow same format we had in python. this branch is directly merging into master. please change it to feature branch.

OptimizelySDK/Utils/Evaluator.cs
```diff
@@ -0,0 +1,102 @@
+﻿/*
```
mikeng13: `Evaluator` is too generic of a name 😄 maybe `ConditionValueEvaluator`?

OptimizelySDK/Utils/Validator.cs
```diff
@@ -114,6 +114,37 @@ public static bool IsUserAttributeValid(KeyValuePair<string, object> attribute)
...
+        /// Validates if the provided value is a valid numeric value.
+        /// </summary>
+        /// <param name="value">Input value</param>
+        /// <returns>true if the provided absolute value is not infinite, NAN and greater than 2^53, false otherwise</returns>
+        public static bool IsValidNumericValue(object value)
+        {
+            if (IsNumericType(value))
+            {
+                var doubleValue = Convert.ToDouble(value);
+                if (double.IsInfinity(doubleValue) || double.IsNaN(doubleValue) || Math.Abs(doubleValue) > Math.Pow(2, 53))
```
mikeng13: I commented on a separate PR for this, but make `Math.Pow(2,53)` into a constant value

PR #118, opened 2018-12-20
feat(project-config): Update audience evaluator and project config for new audience match types.

mfahadahmed:
### Summary

This updates project config and the audience evaluator to finish the implementation of typed audience evaluation:

- In audience_evaluator, allow evaluation to continue when no user attributes are passed (previously we were immediately returning `false` when no attributes were passed)
- Update project_config to update audience_id_map with typed_audiences
- Added unit tests checking that all top-level methods that accept user attributes can bucket users into experiments or rollouts that use typed audiences.

### Test Plan

- New & existing unit tests
- Ran compatibility suite tests for typed audiences

### DIFF
--- a/OptimizelySDK.Tests/OptimizelyTest.cs
+++ b/OptimizelySDK.Tests/OptimizelyTest.cs
@@ -51,6 +51,7 @@ namespace OptimizelySDK.Tests
         private Variation VariationWithKeyControl;
         private Variation VariationWithKeyVariation;
         private Variation GroupVariation;
+        private Optimizely OptimizelyTypedAudience;
 
         #region Test Life Cycle
         [SetUp]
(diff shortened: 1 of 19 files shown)


OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -51,6 +51,7 @@ public class OptimizelyTest
         private Variation VariationWithKeyControl;
         private Variation VariationWithKeyVariation;
         private Variation GroupVariation;
+        private Optimizely OptimizelyTypedAudience;
```
mikeng13: nit: `OptimizelyWithTypedAudiences` to be a little clearer

msohailhussain:
@mikeng13 it's done.

PR #121, opened 2018-12-26
feat(event_builder): Don't track NaN, infinity and value greater than 2^53.

mfahadahmed:
Summary
-------
Don't track NaN, infinity doubles or numeric values greater than 2^53 in EventBuilder.

## Issues

### DIFF
--- a/OptimizelySDK/Utils/Validator.cs
+++ b/OptimizelySDK/Utils/Validator.cs
@@ -23,6 +23,14 @@ namespace OptimizelySDK.Utils
 {
     public static class Validator
     {
+        public static readonly double OPT_NUMBER_LIMIT;
+
+        static Validator()
+        {
+            OPT_NUMBER_LIMIT = Math.Pow(2, 53);
+        }
+
+
         /// <summary>
         /// Validate the ProjectConfig JSON
         /// </summary>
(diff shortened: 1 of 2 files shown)


OptimizelySDK/Utils/Validator.cs
```diff
@@ -23,6 +23,14 @@ namespace OptimizelySDK.Utils
 {
     public static class Validator
     {
+        public static readonly double OPT_NUMBER_LIMIT;
+
+        static Validator()
+        {
+            OPT_NUMBER_LIMIT = Math.Pow(2, 53);
```
alexsaves: Should this not be a `const`?

msohailhussain: No, it can't. Computing values can't be constant.

alexsaves: Why does it have to be computed? 2^53 is `9007199254740992` I think?

`public static const double OPT_NUMBER_LIMIT = 9007199254740992`

alexsaves: Oh, in c# constants can be the result of expressions, I think. So, this would be valid:
`public const int c2 = c1 + 100;`

shouldnt you be able to do `Math.Pow()`? I'm actually curious

msohailhussain: Nope that's not the case, just because of this long value of 2^53 and sync with other SDKs, I added here in this way but changed it now and made comment.

msohailhussain:
done.

PR #122, opened 2019-01-02
feat(audiences): Audience combinations.

mfahadahmed:
Summary
-------
This adds support for audience combinations on experiments. If `experiment.audienceConditions` is present, it will be used as a condition tree where the leaf conditions are audience Ids.

- The audience evaluator uses the condition tree evaluator on both the experiment audience conditions and the individual audience conditions.
- The audience evaluator treats conditions trees and flat lists of audience ids (the old format) the same way, because a flat list of audience ids is a valid condition tree with an implicit OR condition.
- Added a new method to Experiment entity that returns audienceConditions if not null, otherwise returns audience Ids.


OptimizelySDK/Entity/Audience.cs
```diff
@@ -49,7 +50,7 @@ public JToken ConditionList
                 if (DecodedConditions == null)
                 {
                     if (Conditions is string)
-                        DecodedConditions = Utils.ConditionEvaluator.DecodeConditions((string)Conditions);
+                        DecodedConditions = Utils.ConditionTreeEvaluator.DecodeConditions((string)Conditions);
```
msohailhussain: Please rename it with prefix _, private should be with _ at class level variable.

OptimizelySDK/Entity/Experiment.cs
```diff
@@ -69,6 +70,35 @@ public class Experiment : IdKeyEntity
         /// </summary>
         public TrafficAllocation[] TrafficAllocation { get; set; }
 
+        /// <summary>
+        /// Audience Conditions
+        /// </summary>
+        public object AudienceConditions { get; set; }
+
+        private JToken audienceConditionsList = null;
```
msohailhussain: Please use private with _ suffix.

OptimizelySDK/Entity/Experiment.cs
```diff
@@ -147,5 +177,15 @@ public bool IsUserInForcedVariation(string userId)
...
         }
+
+        /// <summary>
+        /// Get audience conditions for the experiment.
+        /// </summary>
+        /// <returns>Audience conditions for the experiment - can be an array of audience IDs, or a nested array of conditions.
+        /// Examples: ["5", "6"], ["and", ["or", "1", "2"], "3"]</returns>
+        public JToken GetAudienceConditionsOrIds()
+        {
+            return AudienceConditionsList != null ? AudienceConditionsList : JArray.FromObject(AudienceIds);
```
msohailhussain: Why you can't use AudienceConditionsList ?? JArray.FromObject(AudienceIds);

msohailhussain:
I have to check this PR in detail.

OptimizelySDK/Utils/ExperimentUtils.cs
```diff
@@ -46,16 +47,34 @@ public static bool IsExperimentActive(Experiment experiment, ILogger logger)
...
 
             if (userAttributes == null)
                 userAttributes = new UserAttributes();
-            
-            var conditionEvaluator = new ConditionEvaluator();
-            return audienceIds.Any(id => conditionEvaluator.Evaluate(config.GetAudience(id).ConditionList, userAttributes).GetValueOrDefault());
+
+            var conditionTreeEvaluator = new ConditionTreeEvaluator();
+
+            bool? EvaluateConditionsWithUserAttributes(JToken condition)
```
msohailhussain: I didn't like inline functions should be written in this way. Please use lambda expression. 
var evaluateConditionsWithUserAttributes = (condition) = > { } should be like this.

msohailhussain:
along with my comments, please use python approach to check validation rather to check in each method.

OptimizelySDK.Tests/UtilsTests/ExperimentUtilsTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            var experiment = Config.GetExperimentFromKey("feat_with_var_test");
+            experiment.AudienceIds = new string[] { };
+            experiment.AudienceConditions = null;
+
+            Assert.True(ExperimentUtils.IsUserInExperiment(Config, experiment, null));
+        }
+
+        [Test]
+        public void TestIsUserInExperimentReturnsFalseWithAudiencesAndNoAttributes()
```
oakbani: This isn't true anymore. We do evaluate even if no attributes are passed. Please revise this to verify that audience get's evaluated with empty/null attributes. 
See this in PHP here: https://github.com/optimizely/php-sdk/blob/master/tests/UtilsTests/ValidatorTest.php#L190

OptimizelySDK.Tests/EntityTests/ExperimentTest.cs
```diff
@@ -0,0 +1,56 @@
...
+        [TestFixtureTearDown]
+        public void Cleanup()
+        {
+            Config = null;
+        }
+
+        #region GetAudienceConditionsOrIds Tests
+
+        [Test]
+        public void TestGetAudienceConditionsOrIdsRetrieveAudienceIdsIfConditionsNotExist()
```
oakbani: Can we also add a testcase with audienceConditions set to empty array?
It should return empty array instead of audienceIds

mfahadahmed: Added this scenario in TestGetAudienceConditionsOrIdsRetrieveAudienceConditionsIfExist test.

OptimizelySDK/Utils/ConditionValueEvaluator.cs
```diff
@@ -95,5 +94,35 @@ public static class AttributeMatchTypes
...
+        /// <returns>true if the type of both values are same, false otherwise.</returns>
+        public static bool AreValuesSameType(object firstValue, object secondValue)
+        {
+            if (firstValue is string && secondValue is string)
+                return true;
+
+            if (firstValue is bool && secondValue is bool)
+                return true;
+
+            if (Validator.IsNumericType(firstValue) && Validator.IsNumericType(secondValue))
```
msohailhussain: ValidNumericType would be more appropriate i guess?

OptimizelySDK/Utils/ExperimentUtils.cs
```diff
@@ -58,18 +58,13 @@ public static bool IsUserInExperiment(ProjectConfig config, Experiment experimen
...
-            }
-
             bool? EvaluateAudience(JToken audienceIdToken)
             {
                 string audienceId = audienceIdToken.ToString();
                 var audience = config.GetAudience(audienceId);
 
                 if (audience != null && !string.IsNullOrEmpty(audience.Id))
-                    return conditionTreeEvaluator.Evaluate(audience.ConditionList, EvaluateConditionsWithUserAttributes);
+                    return conditionTreeEvaluator.Evaluate(audience.ConditionList, condition => CustomAttributeConditionEvaluator.Evaluate(condition, userAttributes));
```
msohailhussain: Please keep this lambda outside and pass the reference variable.

OptimizelySDK/Utils/ConditionTreeEvaluator.cs
```diff
@@ -50,7 +45,7 @@ public class ConditionEvaluator
         /// <param name="userAttributes">Hash representing user attributes</param>
```
mikeng13: Please update the docs with the change in args

OptimizelySDK/Utils/ConditionTreeEvaluator.cs
```diff
@@ -50,7 +45,7 @@ public class ConditionEvaluator
...
         /// <returns>true/false if the user attributes match/don't match the given conditions,
         /// null if the user attributes and conditions can't be evaluated</returns>
-        private bool? AndEvaluator(JArray conditions, UserAttributes userAttributes)
+        private bool? AndEvaluator(JArray conditions, Func<JToken, bool?> leafEvaluator)
```
mikeng13: What is this second function arg we are taking? How come we are using the JToken generic? Can you explain what is going on here?

OptimizelySDK/Utils/CustomAttributeConditionEvaluator.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace OptimizelySDK.Utils
+{
+    public static class CustomAttributeConditionEvaluator
+    {
+        /// <summary>
+        /// String constant representing custome attribute condition type.
+        /// </summary>
+        public const string CUSTOM_ATTRIBUTE_CONDITION_TYPE = "custom_attribute";
+
+        public static bool? Evaluate(JToken condition, UserAttributes userAttributes)
```
mikeng13: Would there be any value in mapping the `condition` to and Optimizely `condition` class instead of passing around the JToken?

OptimizelySDK/Utils/ConditionTreeEvaluator.cs
mfahadahmed: The second argument specify return type of callback which, in this case, is nullable bool for leaf evaluator. Also, we use JToken to represent condition object that is passed to leaf evaluator.

OptimizelySDK/Utils/CustomAttributeConditionEvaluator.cs
mfahadahmed: That's a nice idea but this seems like a major change. Do you want it to be done here in this PR?

mikeng13: No, let's punt on it for now.

PR #127, opened 2019-01-17
ci(codesign): upload build artifacts to s3

juancarlostong:
## Summary
- Update OptimizelySDK.Travis.sln to produce Net40, Net35 builds
- Generate NetStandard16 builds as well (in separate stage)
- Upload dlls to s3
- OASIS-4100

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,14 @@ install:
   - nuget install ./OptimizelySDK.Tests/packages.config -OutputDirectory ./packages
   - nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory ./testrunner
 script:
-  - xbuild /p:Configuration=Release ./OptimizelySDK.Travis.sln
+  - xbuild /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=$TRAVIS_BUILD_DIR/keypair.snk /p:Configuration=Release ./OptimizelySDK.Travis.sln
   - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll
+  # strongname signing from xbuild step needs additional re-signing with sn
+  - find . -path './OptimizelySDK*bin/Release/OptimizelySDK*.dll' -not -regex '.*Tests.*' -print0 | while IFS= read -r -d '' file; do sn -R $file ./keypair.snk; done
 after_success:
   - pip install --user awscli
-  - find . -path './OptimizelySDK*bin/Release/OptimizelySDK*.dll' -not -regex '.*Tests.*' -print0 | while IFS= read -r -d '' file; do (AWS_ACCESS_KEY_ID=$OFTA_KEY AWS_SECRET_ACCESS_KEY=$OFTA_SECRET AWS_DEFAULT_REGION=$OFTA_REGION aws s3 cp $file s3://optly-fs-travisci-artifacts/${TRAVIS_REPO_SLUG}/${TRAVIS_BUILD_NUMBER}/${TRAVIS_JOB_NUMBER}/$(basename $file)-unsigned); done
+  # upload dll build artifacts to s3 with -unsigned suffix
+  - ([ ${TRAVIS_BRANCH} = "master" ] && [ ${TRAVIS_EVENT_TYPE} = "push" ]) && find . -path './OptimizelySDK*bin/Release/OptimizelySDK*.dll' -not -regex '.*Tests.*' -print0 | while IFS= read -r -d '' file; do (AWS_ACCESS_KEY_ID=$OFTA_KEY AWS_SECRET_ACCESS_KEY=$OFTA_SECRET AWS_DEFAULT_REGION=$OFTA_REGION aws s3 cp $file s3://optly-fs-travisci-artifacts/${TRAVIS_REPO_SLUG}/${TRAVIS_BUILD_NUMBER}/${TRAVIS_JOB_NUMBER}/$(basename $file)-unsigned); done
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,3 +37,17 @@ jobs:
       script:
         - "ci/trigger_fullstack-sdk-compat.sh"
       after_success: travis_terminate 0
+    - stage: 'NetStandard16'
+      language: csharp
+      # dotnet only works on trusty https://github.com/travis-ci/travis-ci/issues/5189
+      dist: trusty
+      mono: none
```
juancarlostong: if netstandard16 runs in its own stage i can disable running mono which speeds up build

.travis.yml
```diff
@@ -11,6 +12,11 @@ install:
 script:
   - xbuild /p:Configuration=Release ./OptimizelySDK.Travis.sln
   - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll 
+after_success:
+  - pip install --user awscli
+  - (AWS_ACCESS_KEY_ID=$OFTA_KEY AWS_SECRET_ACCESS_KEY=$OFTA_SECRET AWS_DEFAULT_REGION=$OFTA_REGION aws s3 cp ./OptimizelySDK/bin/Release/OptimizelySDK.dll s3://optly-fs-travisci-artifacts/${TRAVIS_REPO_SLUG}/${TRAVIS_BUILD_NUMBER}/${TRAVIS_JOB_NUMBER}/OptimizelySDK.dll-unsigned)
```
msohailhussain: Can we consolidate above and last 2 lines to single line something like. 
`find ./ -type f | grep pattern | xargs ...`
All dlls will follow the same pattern and also name must follow OptimizelySDK.*.dll

.travis.yml
```diff
@@ -20,7 +24,6 @@ stages:
 jobs:
   include:
     - stage: 'Integration tests'
-      merge_mode: replace
```
juancarlostong: removing this because this command does not take effect (it's not an API call)

.travis.yml
```diff
@@ -10,7 +11,10 @@ install:
   - nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory ./testrunner
 script:
   - xbuild /p:Configuration=Release ./OptimizelySDK.Travis.sln
-  - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll 
+  - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll
```
juancarlostong: whitespace at end

.travis.yml
juancarlostong: fixed with a combination of find and while loop.

.travis.yml
```diff
@@ -10,7 +11,10 @@ install:
...
+after_success:
+  - pip install --user awscli
```
mikeng13: nit: add a comment here saying where we are doing.

I like that we do the install here instead of in the install step because this really is only needed if it succeeds. 

Also, shouldn't we only upload if it is the master branch?

.travis.yml
```diff
@@ -32,3 +35,17 @@ jobs:
       script:
         - "ci/trigger_fullstack-sdk-compat.sh"
       after_success: travis_terminate 0
+    - stage: 'NetStandard16'
```
mikeng13: same question as above, we should only upload if it's the master branch

OptimizelySDK.Travis.sln
```diff
@@ -13,6 +13,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		README.md = README.md
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OptimizelySDK.Net35", "OptimizelySDK.Net35\OptimizelySDK.Net35.csproj", "{C8FF7012-37B7-4D64-AB45-0C62195302EC}"
```
mikeng13: What are these changes for?

OptimizelySDK.Net35/OptimizelySDK.Net35.csproj
```diff
@@ -173,7 +173,7 @@
     <Compile Include="..\OptimizelySDK\Utils\ExperimentUtils.cs">
       <Link>Bucketing\ExperimentUtils</Link>
     </Compile>
-    <Compile Include="..\OptimizelySDK\Bucketing\UserprofileUtil.cs">
+    <Compile Include="..\OptimizelySDK\Bucketing\UserProfileUtil.cs">
```
mikeng13: Oh wow, nice fix. Does this mean that USPs were broken in Net35 and Net40 @msohailhussain ?

msohailhussain: yeah, but fixed here. 
https://github.com/optimizely/csharp-sdk/pull/128/commits/7503fbaf9536819e5dd5a4f10a475a76fa47f00b
When Audience PR will be merged, this will come as conflict.

OptimizelySDK.Travis.sln
juancarlostong: Initially OptimizelySDK.Travis.sln was only using OptimizelySDK\OptimizelySDK.csproj. @msohailhussain added Net35 and Net40 in VS so that those dlls would get generated too. These changes is the result of including those new projects.

PR #132, opened 2019-01-26
fix(eventtagsutils) : don't exclude falsy values

msohailhussain:
## Summary
0 and 1 after deserialization converted into long and we were not considering long in the numeric method. just added it. Along with I realized to add short/uint as well. 
The "why", or other context.

## Issues

### DIFF
--- a/OptimizelySDK.Tests/EventTests/EventBuilderTest.cs
+++ b/OptimizelySDK.Tests/EventTests/EventBuilderTest.cs
@@ -971,7 +971,7 @@ namespace OptimizelySDK.Tests.EventTests
         }
 
         [Test]
-        public void TestConversionEventWithFalsyValues()
+        public void TestConversionEventWithFalsyNumericAndRevenueValues()
         {
             var guid = Guid.NewGuid();
             var timeStamp = TestData.SecondsSince1970();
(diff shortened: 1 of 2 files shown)


OptimizelySDK.Tests/EventTests/EventBuilderTest.cs
```diff
@@ -970,6 +970,183 @@ public void TestConversionEventWithNumericTag()
             Assert.IsTrue(TestData.CompareObjects(expectedEvent, logEvent));
         }
 
+        [Test]
+        public void TestConversionEventWithFalsyValues()
```
mikeng13: nit: with falsy numeric and revenue values

OptimizelySDK.Tests/EventTests/EventBuilderTest.cs
```diff
@@ -970,6 +970,183 @@ public void TestConversionEventWithNumericTag()
...
+                    {"value", 0.0 }
+            });
+
+            TestData.ChangeGUIDAndTimeStamp(logEvent.Params, timeStamp, guid);
+
+            Assert.IsTrue(TestData.CompareObjects(expectedEvent, logEvent));
+        }
+
+        [Test]
+        public void TestConversionEventWithNumericValue1()
```
mikeng13: nit: maybe we should check revenue and event value in separate tests

PR #142, opened 2019-03-11
fix(versioning): NetStandard16

juancarlostong:
## Summary
.csproj for NetStandard16 previously did not include AssemblyInfo.cs
which contains versioning information. Because of this, it used an
automatically generated one similar to the one described here:
https://github.com/dotnet/cli/issues/4783#issuecomment-382695175
We now provide the AssemblyInfo.cs to include during the build and
explicitly tell it not to autogenerate.

## Test plan
after building, check versioning like this:
```
ikdasm -assembly OptimizelySDK.NetStandard16/bin/Debug/netstandard1.6/OptimizelySDK.NetStandard16.dll
Assembly Table
Name:          OptimizelySDK.NetStandard16
Hash Algoritm: 0x00008004
Version:       3.0.0.0
Flags:         0x00000001
PublicKey:     BlobPtr (0x00000e4e)
...
```

### DIFF
--- a/OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
+++ b/OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
@@ -4,6 +4,7 @@
         <TargetFramework>netstandard1.6</TargetFramework>
         <ReleaseVersion>1.2.1</ReleaseVersion>
         <AssemblyOriginatorKeyFile>..\keypair.snk</AssemblyOriginatorKeyFile>
+        <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     </PropertyGroup>
 
     <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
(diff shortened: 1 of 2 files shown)


OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
```diff
@@ -53,6 +54,7 @@
         <Compile Include="..\OptimizelySDK\ProjectConfig.cs" />
         <Compile Include="..\OptimizelySDK\Utils\EventTagUtils.cs" />
         <Compile Include="..\OptimizelySDK\Utils\Validator.cs" />
+        <Compile Include="..\OptimizelySDK\Properties\AssemblyInfo.cs" />
```
msohailhussain: Can you please add one more assemblyinfo file & refer to that file. The problem is AssemblyInfo for .net should be named as OptimizelySDK.NetStandard16 while the referred info has OptimizelySDK name.

OptimizelySDK.NetStandard16/OptimizelySDK.NetStandard16.csproj
```diff
@@ -53,6 +54,7 @@
         <Compile Include="..\OptimizelySDK\ProjectConfig.cs" />
         <Compile Include="..\OptimizelySDK\Utils\EventTagUtils.cs" />
         <Compile Include="..\OptimizelySDK\Utils\Validator.cs" />
+        <Compile Include="..\OptimizelySDK.NetStandard16\Properties\AssemblyInfo.cs" />
```
msohailhussain: Can you try if it works. 
`.\Properties\AssemblyInfo.cs`
instead of 
`..\OptimizelySDK.Net`

PR #149, opened 2019-03-25
feat(api): Decision Notification Listener for IsFeatureEnabled and GetEnabledFeatures APIs.

mfahadahmed:
## Summary
- Implemented Decision Notification Listener for IsFeatureEnabled and GetEnabledFeatures APIs.

## Test plan
- Added unit tests.

### DIFF
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -370,21 +370,45 @@ namespace OptimizelySDK
             if (!Validator.IsFeatureFlagValid(Config, featureFlag))
                 return false;
 
+            string experimentKey = null;
+            string variationKey = null;
+            bool featureEnabled = false;
             var decision = DecisionService.GetVariationForFeature(featureFlag, userId, userAttributes);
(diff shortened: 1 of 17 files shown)


OptimizelySDK/Optimizely.cs
```diff
@@ -645,5 +669,10 @@ private bool ValidateStringInputs(Dictionary<string, string> inputs)
 
             return isValid;
         }
+
+        private bool IsValidDecision(FeatureDecision decision)
```
aliabbasrizvi: What is this validity check about? Also, I feel it belongs on the `FeatureDecision` class and not like this.

aliabbasrizvi: Perhaps call it `IsExperimentDecision`?

mfahadahmed: This check validates decision object. But I misunderstood that DecisionService.GetVariationForFeature() can return null.

I removed this function as the above method can't return null and checking decision.Variation is enough for validating decision object.

PR #150, opened 2019-03-25
feat(api): Decision Notification Listener for feature variable APIs.

mfahadahmed:
## Summary
- Implemented Decision Notification Listener for feature variable APIs.

## Test plan
- Added unit tests.

### DIFF
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -437,7 +437,7 @@ namespace OptimizelySDK
                 featureEnabled = variation.FeatureEnabled.GetValueOrDefault();
                 var featureVariableUsageInstance = variation.GetFeatureVariableUsageFromId(featureVariable.Id);
 
-                if (featureEnabled && featureVariableUsageInstance != null)
+                if (featureVariableUsageInstance != null)
                 {
                     variableValue = featureVariableUsageInstance.Value;
                     Logger.Log(LogLevel.INFO,
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -371,7 +371,7 @@ namespace OptimizelySDK
                 return false;
 
             var decision = DecisionService.GetVariationForFeature(featureFlag, userId, userAttributes);
-            if (IsValidDecision(decision)) {
+            if (decision.Variation != null) {
                 if (decision.Source == FeatureDecision.DECISION_SOURCE_EXPERIMENT) {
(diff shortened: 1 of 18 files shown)


OptimizelySDK/Optimizely.cs
```diff
@@ -655,5 +646,33 @@ private bool ValidateStringInputs(Dictionary<string, string> inputs)
...
+        {
+            object result = null;
+            switch (type)
+            {
+                case FeatureVariable.VariableType.BOOLEAN:
+                    bool.TryParse(value, out bool booleanValue);
+                    result = booleanValue;
+                    break;
+                case FeatureVariable.VariableType.DOUBLE:
+                    double.TryParse(value, out double doubleValue);
```
mikeng13: Does it matter that in the previous code we used the `Double` class to parse and here we are calling it from the `double` primitive type? 

`Double.TryParse(variableValue, out double doubleValue)`

OptimizelySDK/Optimizely.cs
```diff
@@ -655,5 +646,33 @@ private bool ValidateStringInputs(Dictionary<string, string> inputs)
...
+                    result = doubleValue;
+                    break;
+                case FeatureVariable.VariableType.INTEGER:
+                    int.TryParse(value, out int intValue);
```
mikeng13: Same question as above, where we use to parse from the `Int32` class

OptimizelySDK/Optimizely.cs
```diff
@@ -406,7 +406,7 @@ public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttri
         /// <param name="userAttributes">The user's attributes</param>
         /// <param name="variableType">Variable type</param>
         /// <returns>string | null Feature variable value</returns>
-        public virtual string GetFeatureVariableValueForType(string featureKey, string variableKey, string userId, 
+        public virtual T GetFeatureVariableValueForType<T>(string featureKey, string variableKey, string userId,
```
mikeng13: This is an interesting improvement :)

OptimizelySDK/Optimizely.cs
msohailhussain: https://stackoverflow.com/questions/3951224/is-there-a-meaningful-difference-between-double-and-double-in-net
No difference.

OptimizelySDK/Optimizely.cs
msohailhussain: yeah, we actually needed to notify for every featurevariable* api, and type was required, thats why used generics here.

PR #157, opened 2019-04-22
feat(decision-listener): Incorporated new decision notification listener changes.

mfahadahmed:
## Summary
- Incorporated new decision notification listener changes.

## Test plan
- Added unit tests.

### DIFF
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -308,7 +308,8 @@ namespace OptimizelySDK
             };
 
             userAttributes = userAttributes ?? new UserAttributes();
-            NotificationCenter.SendNotifications(NotificationCenter.NotificationType.Decision, DecisionInfoTypes.EXPERIMENT, userId,
+            var decisionNotificationType = Config.IsFeatureExperiment(experiment.Id) ? DecisionNotificationTypes.FEATURE_TEST : DecisionNotificationTypes.AB_TEST;
+            NotificationCenter.SendNotifications(NotificationCenter.NotificationType.Decision, decisionNotificationType, userId,
                 userAttributes, decisionInfo);
             return variation;
         }
@@ -380,9 +381,8 @@ namespace OptimizelySDK
             if (!Validator.IsFeatureFlagValid(Config, featureFlag))
                 return false;
 
-            string experimentKey = null;
-            string variationKey = null;
             bool featureEnabled = false;
+            var sourceInfo = new Dictionary<string, string>();
             var decision = DecisionService.GetVariationForFeature(featureFlag, userId, userAttributes);
 
             if (decision.Variation != null)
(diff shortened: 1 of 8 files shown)


mikeng13:
Can you please link to a custom build that runs the compat suite tests for this?

mfahadahmed:
compat suite custom build
https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/109547543

OptimizelySDK/ProjectConfig.cs
```diff
@@ -617,5 +635,15 @@ public string GetAttributeId(string attributeKey)
             Logger.Log(LogLevel.ERROR, $@"Attribute key ""{attributeKey}"" is not in datafile.");
             return null;
         }
+
+        /// <summary>
+        /// Check if the provided experiment Id belongs to any feature, false otherwise.
```
mikeng13: We should add a unit test for this method with the `ExperimentFeatureMap` being populated

mfahadahmed: Unit tests added for `IsFeatureExperiment` method.

OptimizelySDK/ProjectConfig.cs
```diff
@@ -293,6 +299,18 @@ private void Initialize()
...
+
+            // Adding experiments in experiment-feature map.
+            foreach (var feature in FeatureFlags)
+            {
+                foreach (var experimentId in feature.ExperimentIds ?? new List<string>())
+                {
+                    if (ExperimentFeatureMap.ContainsKey(experimentId))
+                        ExperimentFeatureMap[experimentId].Add(feature.Id);
+                    else
+                        ExperimentFeatureMap[experimentId] = new List<string> { experimentId };
```
aliabbasrizvi: Shouldn't we be having `feature.Id` and not `experimentId`?

mfahadahmed: Oops, yeah that's a mistake. Fixed now.

aliabbasrizvi:
@mikeng13 can you take a look as well?

OptimizelySDK/Optimizely.cs
```diff
@@ -308,7 +308,8 @@ public Variation GetVariation(string experimentKey, string userId, UserAttribute
             };
 
             userAttributes = userAttributes ?? new UserAttributes();
-            NotificationCenter.SendNotifications(NotificationCenter.NotificationType.Decision, DecisionInfoTypes.EXPERIMENT, userId,
+            var decisionNotificationType = Config.IsFeatureExperiment(experiment.Id) ? DecisionNotificationTypes.FEATURE_TEST : DecisionNotificationTypes.AB_TEST;
```
msohailhussain: Can we change it instead of returning `true` / `false`, we should return directly the DecisionNotificationTypes.
Because I assume, no need to create a function to check only Key exist in dictionary or not, which can directly do it here. 
@aliabbasrizvi @mikeng13

mikeng13: I think this is fine. No need to expose the map from the project config object if we don't have to. And it's consistent with what we did on the JS SDK.

PR #160, opened 2019-05-29
feat(dfm): Datafile management implementation

msohailhussain:
## Summary
All details are in the doc. 

## Issues


OptimizelySDK/DatafileManagement/HttpProjectConfigManager.cs
```diff
@@ -0,0 +1,237 @@
...
+        public Task OnReady()
+        {
+            return CompletableConfigManager.Task;
+            //Task t = new Task(() => {
+            //    CompletableConfigManager.Task.Wait(this.BlockingTimeout);
+            //});
+            //t.Start();
+            //return t;
+        }
+        static ProjectConfig ParseProjectConfig(string datafile)
```
mikeng13: I don't think this is the right place to put it. If you look at the Java SDK, the notion of a ProjectConfig backed by a Datafile is split from this and is defined at a higher level: https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/config/DatafileProjectConfig.java

The idea is that we can have other types of Datafile-backed project config that don't necessarily have to be polled, such as push-based with websockets or server-sent events.

OptimizelySDK/DatafileManagement/ProjectConfig.cs
```diff
@@ -0,0 +1,269 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using OptimizelySDK.Entity;
+using System.Collections.Generic;
+
+namespace OptimizelySDK
+{
+    public interface ProjectConfig
```
mikeng13: ProjectConfig can be more abstract than this. We can have project configs later on that are not based on datafile, but can be based on a different structure. Here is the abstraction approach we took in the Java SDK: 

https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java

For this reason, also, we should change the name of the package to `ProjectConfig` instead of `DatafileManagement` so we don't tie it to just the Datafile transport type.

OptimizelySDK/Config/DatafileProjectConfig.cs
```diff
@@ -0,0 +1,643 @@
+﻿/* 
+ * Copyright 2017-2019, Optimizely
```
mikeng13: nit: Should just be 2019

OptimizelySDK/Config/DatafileProjectConfig.cs
```diff
@@ -0,0 +1,643 @@
...
+        private Dictionary<string, Audience> _AudienceIdMap;
+        public Dictionary<string, Audience> AudienceIdMap { get { return _AudienceIdMap; } }
+
+        /// <summary>
+        /// Associative array of user IDs to an associative array
+        /// of experiments to variations.This contains all the forced variations
+        /// set by the user by calling setForcedVariation (it is not the same as the
+        /// whitelisting forcedVariations data structure in the Experiments class).
+        /// </summary>
+        private Dictionary<string, Dictionary<string, string>> _ForcedVariationMap;
```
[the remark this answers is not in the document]
mikeng13: +1

OptimizelySDK/Config/DatafileProjectConfig.cs
```diff
@@ -0,0 +1,643 @@
...
+        public Dictionary<string, Rollout> RolloutIdMap { get { return _RolloutIdMap; } }
+
+        /// <summary>
+        /// Associative array of experiment IDs that exist in any feature
+        /// for checking that experiment is a feature experiment.
+        /// </summary>
+        private Dictionary<string, List<string>> ExperimentFeatureMap = new Dictionary<string, List<string>>();
+
+
+        //========================= Callbacks ===========================
```
mikeng13: nit: These aren't really callbacks are they?

OptimizelySDK/Config/FallbackProjectConfigManager.cs
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OptimizelySDK.Config
+{
+    public class FallbackProjectConfigManager : ProjectConfigManager
```
mikeng13: Is there gonna be more logic for the "fallback"? I think you mentioned you'd have that in a separate PR. In that case just add a `//TODO` here for it

OptimizelySDK/Config/HttpProjectConfigManager.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        private HttpProjectConfigManager(TimeSpan period, string url, TimeSpan blockingTimeout, ILogger logger, IErrorHandler errorHandler) : base(period, blockingTimeout, logger, errorHandler)
+        {
+            Url = url;
+        }
+
+        public Task OnReady()
+        {
+            return CompletableConfigManager.Task;
+            //Task t = new Task(() => {
```
mikeng13: nit: remove

OptimizelySDK/Config/DatafileProjectConfig.cs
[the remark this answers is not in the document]
msohailhussain: It's been moved in sohail/optimizelydfm (Will add PR today)

OptimizelySDK/Config/FallbackProjectConfigManager.cs
msohailhussain: nope, no more logic, in other PR, I will remov SetConfig, and ProjectConfig will only be set by constructor.

msohailhussain:
> lgtm. We just need to remove the `ForcedVariationMap` from the `DatafileProjectConfig`

I removed in this PR.
https://github.com/optimizely/csharp-sdk/pull/161/files#diff-6c7503ba59bde21387a0424b504ad19dL137

PR #161, opened 2019-06-04
feat(dfm): Integrate DFM with optimizely

msohailhussain:
## Summary
- Added ProjectConfigManager in Optimizely class
- Added OptimizelyFactory
- Inside every API, replaced config with ProjectConfig Manager
- Removed ProjectConfig from constructor
- DecisionService using ProjectConfig with every method. 
- Removed ForcedVariation from ProjectConfig & DatafileProjectConfig
- Moved ForcedVariation to DecisionService
- Added GetVariation private method to use ProjectConfig given by Activate. 
- Updated existing unit tests

TODO:
- Resolve TODOs (Mostly documentation & unit tests to be added)

## Issues


msohailhussain:
OK, I will separate out ForcedVariation, infact @mfahadahmed will do it.

OptimizelySDK/Config/FallbackProjectConfigManager.cs
```diff
@@ -16,20 +16,31 @@
 
 namespace OptimizelySDK.Config
 {
-    /// TODO: Cleanup done in other PR. sohail/optimizelydfm
+    /// <summary>
+    /// Implementation of ProjectConfigManager interface that simply
+    /// returns the stored ProjectConfig instance.
+    /// </summary>
```
mikeng13: nit: Also mention that it is static and the config it holds cannot change

OptimizelySDK/Optimizely.cs
```diff
@@ -129,14 +118,47 @@ public class Optimizely : IOptimizely
...
+                         IEventDispatcher eventDispatcher = null,
+                         ILogger logger = null,
+                         IErrorHandler errorHandler = null,
+                         UserProfileService userProfileService = null)
+        {
+            ProjectConfigManager = configManager;
+            Initialize(eventDispatcher, logger, errorHandler, userProfileService);
+        }
+
+        private void Initialize(IEventDispatcher eventDispatcher = null,
```
mikeng13: nit: `InitializeComponents`?

OptimizelySDK/Optimizely.cs
```diff
@@ -367,6 +415,14 @@ public Variation GetForcedVariation(string experimentKey, string userId)
         /// <returns>True if feature is enabled, false or null otherwise</returns>
         public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttributes userAttributes = null)
         {
+            var config = ProjectConfigManager?.GetConfig();
+            if (!IsValid && config == null) {
+
+                Logger.Log(LogLevel.ERROR, "Datafile has invalid format. Failing 'GetVariation'.");
```
mikeng13: `... Failing 'IsFeatureEnabled'`

OptimizelySDK/Optimizely.cs
```diff
@@ -529,7 +586,14 @@ public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttri
...
         {
-            return GetFeatureVariableValueForType<bool?>(featureKey, variableKey, userId, userAttributes, FeatureVariable.VariableType.BOOLEAN);
+            var config = ProjectConfigManager?.GetConfig();
+
+            if (!IsValid && config == null) {
+                Logger.Log(LogLevel.ERROR, "Datafile has invalid format. Failing 'activate'.");
+                return null;
+            }
+
+            return GetFeatureVariableValueForType<bool?>(featureKey, variableKey, userId, userAttributes, FeatureVariable.VariableType.BOOLEAN, config);
```
mikeng13: Why do we need to pass the config to it? Why not have the `GetFeatureVariableValueForType` method use the `ProjectConfigManager`?

OptimizelySDK/Optimizely.cs
```diff
@@ -529,7 +586,14 @@ public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttri
         /// <returns>bool | Feature variable value or null</returns>
         public bool? GetFeatureVariableBoolean(string featureKey, string variableKey, string userId, UserAttributes userAttributes = null)
         {
-            return GetFeatureVariableValueForType<bool?>(featureKey, variableKey, userId, userAttributes, FeatureVariable.VariableType.BOOLEAN);
+            var config = ProjectConfigManager?.GetConfig();
+
+            if (!IsValid && config == null) {
+                Logger.Log(LogLevel.ERROR, "Datafile has invalid format. Failing 'activate'.");
```
mikeng13: Not `activate` but `GetFeatureVariableBoolean`

OptimizelySDK/Optimizely.cs
```diff
@@ -568,22 +644,28 @@ public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttri
...
         /// </summary>
         /// <param name="experiment">The experiment</param>
-        /// <param name="variationId">The variation entity</param>
+        /// <param name="variation">The variation entity</param>
         /// <param name="userId">The user ID</param>
         /// <param name="userAttributes">The user's attributes</param>
-        private void SendImpressionEvent(Experiment experiment, Variation variation, string userId, 
-        UserAttributes userAttributes)
+        private void SendImpressionEvent(Experiment experiment, Variation variation, string userId,
+                                         UserAttributes userAttributes, ProjectConfig config)
```
mikeng13: We don't need to take in the `ProjectConfig` as a param. We can use the `ProjectConfigManager` in here

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+using OptimizelySDK.Bucketing;
+using OptimizelySDK.Config;
+using OptimizelySDK.ErrorHandler;
+using OptimizelySDK.Event.Dispatcher;
+using OptimizelySDK.Logger;
+
+namespace OptimizelySDK
+{
+
+    /// TODO: Need to check with Mike, no need of this class
```
mikeng13: This is not the case anymore right?

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+            return NewDefaultInstance(sdkKey, null);
+        }
+
+        public static Optimizely NewDefaultInstance(string sdkKey, string fallback)
+        {
+            var logger = new DefaultLogger();
+            var errorHandler = new DefaultErrorHandler();
+            var eventDispatcher = new DefaultEventDispatcher(logger);
+            var builder = new HttpProjectConfigManager.Builder();
+            // don't block
```
mikeng13: What is this doc comment for?

OptimizelySDK/Optimizely.cs
```diff
@@ -168,7 +190,8 @@ private bool ValidatePreconditions(Experiment experiment, string userId, UserAtt
         /// <returns>null|Variation Representing variation</returns>
         public Variation Activate(string experimentKey, string userId, UserAttributes userAttributes = null)
         {
-            if (!IsValid)
+            var config = ProjectConfigManager?.GetConfig();
+            if (!IsValid && config == null)
```
mjc1283: Should this (and the same check elsewhere) be using an or operator instead of and operator? I.e., `!IsValid || config == null`

OptimizelySDK/Optimizely.cs
msohailhussain: This method is used by `Activate`, if we use `ProjectConfigManager` inside `SendImpressionEvent`, there are chances `ProjectConfig` used to get `variation` from `DecisionService` may not be same when get from `ProjectConfigManagr` inside this method.

OptimizelySDK/OptimizelyFactory.cs
msohailhussain: Yes, will add unit test and more convenience methods in separate PR. Removing this comment.

OptimizelySDK/Optimizely.cs
msohailhussain: It should be AND, if isValid is false and also Config is null, then it's invalid.
I will consolidate these conditions into one.

OptimizelySDK/Optimizely.cs
mikeng13: Hmm, ok that could actually happen.

OptimizelySDK/Optimizely.cs
```diff
@@ -488,9 +488,15 @@ public virtual bool IsFeatureEnabled(string featureKey, string userId, UserAttri
         /// <param name="variableType">Variable type</param>
         /// <returns>string | null Feature variable value</returns>
         public virtual T GetFeatureVariableValueForType<T>(string featureKey, string variableKey, string userId, 
-            UserAttributes userAttributes, FeatureVariable.VariableType variableType, ProjectConfig config)
+            UserAttributes userAttributes, FeatureVariable.VariableType variableType)
         {
 
+            var config = ProjectConfigManager?.GetConfig();
+            if (!IsValid && config == null) {
+                Logger.Log(LogLevel.ERROR, "Datafile has invalid format. Failing 'activate'.");
```
mikeng13: This still refers to `activate`. Please make sure it references the right API name.

OptimizelySDK/Config/FallbackProjectConfigManager.cs
```diff
@@ -18,7 +18,7 @@ namespace OptimizelySDK.Config
 {
     /// <summary>
     /// Implementation of ProjectConfigManager interface that simply
-    /// returns the stored ProjectConfig instance.
+    /// returns the stored ProjectConfig instance which is mutable and cannot be changed.
```
mikeng13: Hmm this is confusing. The `ProjectConfig` in this manager should not be mutable.

msohailhussain:
@mikeng13 Please check my last commit, feedback addressed.

OptimizelySDK/Optimizely.cs
mjc1283: `ProjectConfigManager.GetConfig` can return `null`. When it does, we must return `null` early, otherwise attempting to access `config.GetExperimentFromKey` could crash. If I'm misunderstanding, let me know.

msohailhussain: So `IsValid` will always be `false` if we are using `ProjectConfigManager` to initialize `Optimizely` instance. 
Please let me know we can chat for more clarity.

mikeng13: Since we are opening up the `ProjectConfigManager` to be overriden, we should guard against the custom implementation not returning `null` or not setting `isValid` to false. Therefore I think the right apporach is to be more careful and use `!IsValid || config == null` like @mjc1283 suggested.

PR #163, opened 2019-06-14
feat(OptimizelyConfigUPdate): Notification config update

msohailhussain:
## Summary
- Added one more Notification type for config update
- Added event handler for emitting notification from PollingProjectConfigManager
- autoUpdate was missing, added in this PR
- Using interlocked, instead of Monitor due to its limitation with `Task/await`.

The "why", or other context.

## Issues


msohailhussain:
https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/jobs/208722466

OptimizelySDK/Config/HttpProjectConfigManager.cs
```diff
@@ -182,6 +183,20 @@ public Builder WithErrorHandler(IErrorHandler errorHandler)
...
             }
 
+            public Builder WithAutoUpdate(bool autoUpdate)
+            {
+                AutoUpdate = autoUpdate;
+
+                return this;
+            }
+
+            public Builder WithStartByDefault(bool startByDefault=true)
```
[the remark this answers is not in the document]
msohailhussain: @mjc1283 @mikeng13 You are right but I added `Start` because there are chances to lost first `Notification`. Even if you `Start` by default `HttpProjectConfigManager`, it emits event before subscribing and initializing `Optimizely` instance.
To overcome this issue, I think `NotificationCenter` should be passed to `HttpProjectConfigManager`, like implemented in Java. If you agree I can make this change?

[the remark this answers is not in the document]
mjc1283: That change sounds good to me. Thanks.

OptimizelySDK/Config/PollingProjectConfigManager.cs
```diff
@@ -34,33 +34,29 @@ public abstract class PollingProjectConfigManager : ProjectConfigManager
...
         protected TaskCompletionSource<bool> CompletableConfigManager = new TaskCompletionSource<bool>();
         // Variables to control blocking/syncing.
-        public object mutex = new Object();
+        public int resourceInUse = 0;
 
-        protected event Action ProjectConfig_Notification;
+        public event Action NotifyOnProjectConfigUpdate;
 
-        public PollingProjectConfigManager(TimeSpan period, TimeSpan blockingTimeout, ILogger logger = null, IErrorHandler errorHandler = null, bool StartByDefault = true)
+        public PollingProjectConfigManager(TimeSpan period, TimeSpan blockingTimeout, bool autoUpdate = false, ILogger logger = null, IErrorHandler errorHandler = null)
```
mjc1283: `autoUpdate` defaults to `true` in server-side JS, and I believe Java also.

msohailhussain:
Yes, changing Synchronization in Run method is because Monitor.Exit, only exits in the same thread where it was acquired. We are using Poll method which may use Task, and Task may change Thread, that's why I used Interlocked which is recommended in our case.

OptimizelySDK/Config/PollingProjectConfigManager.cs
msohailhussain: Sure, I will make this change and update it.

OptimizelySDK/Config/PollingProjectConfigManager.cs
```diff
@@ -154,20 +150,24 @@ public bool SetConfig(ProjectConfig projectConfig)
...
+        }
 
         /// <summary>
         /// Scheduler method that periodically runs on provided
         /// polling interval.
         /// </summary>
         public virtual void Run()
         {
-            if (Monitor.TryEnter(mutex)){
+            if (Interlocked.Exchange(ref resourceInUse, 1) == 0){
```
mikeng13: This is pretty cool.

msohailhussain:
latest build
https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/116231988

OptimizelySDK/Config/PollingProjectConfigManager.cs
```diff
@@ -150,9 +150,13 @@ public bool SetConfig(ProjectConfig projectConfig)
             CurrentProjectConfig = projectConfig;
 
             // SetResult raise exception if called again, that's why Try is used.
-            NotifyOnProjectConfigUpdate?.Invoke();
             CompletableConfigManager.TrySetResult(true);
 
+            // Invoke event in separate thread.
```
mikeng13: nit: Separate task? Also please explain why this is necessary in the comment.

PR #170, opened 2019-06-20
feat(dispose): Added Disposable interface

msohailhussain:
## Summary
- The "what"; a concise description of each logical change
- Another change

The "why", or other context.

## Issues


OptimizelySDK/Optimizely.cs
```diff
@@ -781,5 +782,14 @@ private object GetTypeCastedVariableValue(string value, FeatureVariable.Variable
...
         }
-    }
+
+        public void Dispose()
+        {
+            if (Disposed) return;
+
+            Disposed = true;
+            (ProjectConfigManager as IDisposable)?.Dispose();
+            ProjectConfigManager = null;
```
mikeng13: If we dispose the optimizely client, but then accidentally end up using it again afterwards, will this result in a null pointer exception?

msohailhussain: Added test case here. 
https://github.com/optimizely/csharp-sdk/pull/170/files#diff-8660e0a3c8615b111f08b9e4ab4f9b36R3485

mikeng13: 👍

mikeng13:
@msohailhussain the unit test failure seems legit

PR #171, opened 2019-07-03
feat(epmodel): Event Processor datamodel

msohailhussain:
## Summary
- Data models to represent Impression/Conversion event on client/server side

## Test plan
- Unit test to verify after serialization datamodel is converted into expected server payload
  - Conversion Event (Already there)
 - Impression Event (In Progress)

## Issues


msohailhussain:
@mikeng13 Hey Mike, yeah that was a mistake, corrected it. 

But I am not 100% convinced to have a `EventContext` as reference instead of inheriting. May be we may discuss offline. Also I have removed the unit-test which I was thinking `EventBatch` to stick with the expected payload while serializing into JSON.

PR #172, opened 2019-07-10
fix(defaulttiming): issues fixed related to default timings

msohailhussain:
## Summary
Polling interval should be default when 0 or -ve value provided
Blocking timeout should be default when 0 or -ve value provided
Due call, should be made after releasing lock. 
The "why", or other context.

## Issues

### DIFF
--- a/OptimizelySDK/Config/HttpProjectConfigManager.cs
+++ b/OptimizelySDK/Config/HttpProjectConfigManager.cs
@@ -122,14 +122,18 @@ namespace OptimizelySDK.Config
         
         public class Builder
         {
+            private const long MAX_MILLISECONDS_LIMIT = 4294967294;
+            private readonly TimeSpan DEFAULT_PERIOD = TimeSpan.FromMinutes(5);
+            private readonly TimeSpan DEFAULT_BLOCKINGOUT_PERIOD = TimeSpan.FromSeconds(15);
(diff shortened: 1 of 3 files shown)


OptimizelySDK/Config/HttpProjectConfigManager.cs
```diff
@@ -122,14 +122,18 @@ protected override ProjectConfig Poll()
         
         public class Builder
         {
+            private const long MAX_MILLISECONDS_LIMIT = 4294967294;
```
mikeng13: Cool, we need to make sure that we document these values for options in the README in your other PR

msohailhussain: > Are these compat suite tests that we can include with this fix to ensure that other SDKs aren't similarly affected?

Sure, we can add FSC test cases.

msohailhussain:
https://github.com/optimizely/csharp-sdk/pull/172#pullrequestreview-260333443
Sure, will add FSC test cases for these scenarios

OptimizelySDK/Config/HttpProjectConfigManager.cs
msohailhussain: Sure, will add this value in README.md

PR #174, opened 2019-07-11
feat(eventfactory): Event Factory and UserEventFactory

msohailhussain:
## Summary
- UserEventFactory decouples ProjectConfig & provide the object to represent impression and conversion event.
- EventFactory represents EventBatch after grouping imp/conv objects.

## Test plan
- Converted EventBuilder unit tests to EventFactory to make sure, all scenarios are covered to send payload to Server.


OptimizelySDK/Event/Entity/SnapshotEvent.cs
```diff
@@ -36,12 +39,12 @@ public class SnapshotEvent
...
         [JsonProperty("value", NullValueHandling = NullValueHandling.Ignore)]
-        public long? Value { get; private set; }
+        public float? Value { get; private set; }
 
         [JsonProperty("tags", NullValueHandling = NullValueHandling.Ignore)]
         public EventTags EventTags { get; private set; }
-
-        public SnapshotEvent(string entityId, string uuid, string key, long timestamp, int? revenue, long? value, EventTags eventTags)
+        
+        public SnapshotEvent(string entityId, string uuid, string key, long timestamp, int? revenue = null, float? value = null, EventTags eventTags = null)
```
mikeng13: How come `SnapshotEvent` doesn't get a `Builder`?

OptimizelySDK/Event/Entity/UserEvent.cs
```diff
@@ -13,12 +13,28 @@
...
+    /// </summary>
     public class UserEvent
     {
-        public EventContext Context { get; private set; }
-        public string UUID { get; private set; }
-        public long TimeStamp { get; private set; }
+        /// <summary>
+        /// Helper to compute Unix time (i.e. since Jan 1, 1970)
+        /// </summary>
+        protected static long SecondsSince1970
```
mikeng13: This seems like it could be a more global utility

OptimizelySDK/Event/Internals/EventFactory.cs
```diff
@@ -0,0 +1,200 @@
+﻿/* 
+ * Copyright 2017-2019, Optimizely
```
mikeng13: Only 2019

OptimizelySDK/Event/Internals/EventFactory.cs
```diff
@@ -0,0 +1,200 @@
+﻿/*
```
mikeng13: I don't think there is a need for making things `internal` anymore. We want to be able to expose as much of the SDK as possible. Plus the class itself it public and not protected :)

OptimizelySDK/Event/Internals/UserEventFactory.cs
```diff
@@ -0,0 +1,98 @@
+﻿using OptimizelySDK.Entity;
+using OptimizelySDK.Event.Entity;
+
+namespace OptimizelySDK.Event.internals
+{
+    /// <summary>
+    /// UserEventFactory builds ImpressionEvent and ConversionEvent objects from a given UserEvent.
+    /// </summary>
+    public class UserEventFactory
```
mikeng13: Is there a strong reason for making this a separate file? I was thinking a generic "Event" factory can create all these types of events

OptimizelySDK/Event/Entity/SnapshotEvent.cs
msohailhussain: you are right, I will add builder for it, initially I decided not to have builder for simple classes but this class have same type of properties which may be interchanged. I actually missed it. Will add

OptimizelySDK/Event/Internals/EventFactory.cs
msohailhussain: I will remove it from `Internals`, just FYI, class should have `internal` attribute not to expose outside of assembly.

OptimizelySDK/Event/Internals/UserEventFactory.cs
msohailhussain: I think it's a categorization between userevents which are impression and conversion and server event which is only LogEvent. Same in java. that's why I did it.

mikeng13: Ok that is fine then.

OptimizelySDK/Utils/GeneralUtils.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
mikeng13: This can be less generic. We can call it `DateUtils`

PR #191, opened 2019-08-01
feat(eventprocess): Integrate Event Processor with Optimizely

msohailhussain:
## Summary
- Integrated Event Processor with Optimizely class
  - Added argument in the constructor for EP
- Default ForwardingEventProcessor is implemented and used while not providing EP

## Test plan
- Added Unit tests
- Need to test with FSC.

## Issues
- Need to remove all warnings. 
- Remove all deprecated methods of notification center.


OptimizelySDK/Optimizely.cs
```diff
@@ -293,23 +299,16 @@ public void Track(string eventKey, string userId, UserAttributes userAttributes
...
-                EventDispatcher.DispatchEvent(conversionEvent);
-            }
-            catch (Exception exception)
+            if (NotificationCenter.GetNotificationCount(NotificationCenter.NotificationType.Track) > 0)
             {
-                Logger.Log(LogLevel.ERROR, string.Format("Unable to dispatch conversion event. Error {0}", exception.Message));
-            }
-
-            NotificationCenter.SendNotifications(NotificationCenter.NotificationType.Track, eventKey, userId,
+                var conversionEvent = EventFactory.CreateLogEvent(userEvent, Logger);
```
mikeng13: Hmm if we pass it a logger, wouldn't it be weird that the same message gets logged twice because we also create a log event inside the processor. Why not have the forwarding event processor also take in the notification center and send the notification for the notification events? Seems less duplicative than creating the log events multiple times

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -58,8 +77,17 @@ public static Optimizely NewDefaultInstance(string sdkKey, string fallback)
 
         public static Optimizely NewDefaultInstance(ProjectConfigManager configManager, NotificationCenter notificationCenter = null, IEventDispatcher eventDispatcher = null,
                                                     IErrorHandler errorHandler = null, ILogger logger = null, UserProfileService userprofileService = null)
-        {            
-            return new Optimizely(configManager, notificationCenter, eventDispatcher, logger, errorHandler, userprofileService);      
+        {
+            EventProcessor eventProcessor = null;
+
+#if !NETSTANDARD1_6
+            eventProcessor = new BatchEventProcessor.Builder()
```
mikeng13: This guy needs to take in the notification center.

OptimizelySDK/Optimizely.cs
```diff
@@ -666,22 +668,19 @@ public string GetFeatureVariableString(string featureKey, string variableKey, st
...
+                EventProcessor.Process(userEvent);
                 Logger.Log(LogLevel.INFO, string.Format("Activating user {0} in experiment {1}.", userId, experiment.Key));
-                Logger.Log(LogLevel.DEBUG, string.Format("Dispatching impression event to URL {0} with params {1}.",
-                    impressionEvent.Url, impressionEvent.GetParamsAsJson()));
 
-                try
+                // Kept For backwards compatibility.
+                // This notification is deprecated and the new DecisionNotifications
+                // are sent via their respective method calls.
+                if (NotificationCenter.GetNotificationCount(NotificationCenter.NotificationType.Activate) > 0)
```
aliabbasrizvi: This check does not look great. Should we not try to call this method and this becomes a no-op if there are no listeners like earlier?

OptimizelySDK/Event/BatchEventProcessor.cs
```diff
@@ -251,12 +253,15 @@ private void AddToBatch(UserEvent userEvent)
         public void Dispose()
         {
             if (Disposed) return;
+
+            Stop();
```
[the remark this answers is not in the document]
msohailhussain: I have added unit tests, that make sure, IsStarted become false on Dispose.

OptimizelySDK/Event/ForwardingEventProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+            ErrorHandler = errorHandler ?? new DefaultErrorHandler(Logger, false);
+        }
+
+        public void Process(UserEvent userEvent)
+        {
+            var logEvent = EventFactory.CreateLogEvent(userEvent, Logger);            
+
+            try
+            {
+                if (userEvent is ImpressionEvent)
```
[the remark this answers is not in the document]
msohailhussain: removed logging, logging params when exception is raised.

OptimizelySDK/Optimizely.cs
msohailhussain: I really like your suggestion but Activate and Track are already obsoleted, this condition will be removed in the upcoming versions and also we are overriding delegates, doesn't have the same pattern we have in java.

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -32,6 +33,24 @@ namespace OptimizelySDK
     /// TODO: Add unit test of this class.
     public static class OptimizelyFactory
     {
+        // @sohail: We can also create a new static method NewDefaultInstance(batchsize, flushInterval = null)
```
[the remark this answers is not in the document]
msohailhussain: removed. thanks

PR #197, opened 2019-09-02
Feat: Add Config support for initializing EP & HttpProjectConfigManager

mnoman09:
## Summary
- Added new method of NewDefaultInstance taking no parameters 
- This new method will take all the available configuration from app.config and initialize EventProcessor and HttpProjectConfigManager

## Test plan
- Added Unit Tests 

## Ticket
- OASIS-5146


OptimizelySDK.Tests/App.config
```diff
@@ -4,10 +4,31 @@
...
-    </appSettings>
+  <configSections>
+    <section name="optlySDKConfigSection"
+             type="OptimizelySDK.Config.OptimizelySDKConfigSection, OptimizelySDK, Version=3.2.0.0, Culture=neutral, PublicKeyToken=null" />
+  </configSections>
+  
+  <optlySDKConfigSection>
+  
+    <HttpProjectConfig sdkKey="43214321" 
+                       url="www.testurl.com"
```
mikeng13: This url is not needed because you have the url format

OptimizelySDK.Tests/App.config
```diff
@@ -4,10 +4,31 @@
     that apply only to the Test project.
 -->
 <configuration>
-    <appSettings>
 
-    </appSettings>
+  <configSections>
```
mikeng13: Can you please add notes in the README for this and how it affects unit tests?

OptimizelySDK/Config/OptimizelyConfigHandler.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System.Configuration;
```
mikeng13: Needs license header. Also, can you add a doc description for what this class is for? 

Also, if this also has config options for the event processor, it should not be in the config package. Instead it should be on the same level as the client. We don't need to have `Optimizely` in the name. Instead it can be `ClientConfigHandler`

OptimizelySDK/Config/OptimizelyConfigHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            get { return (string)base["url"]; }
+        }
+
+        [ConfigurationProperty("datafileUrlFormat")]
+        public string DatafileUrlFormat
+        {
+            get { return (string)base["datafileUrlFormat"]; }
+        }
+
+        [ConfigurationProperty("pollingIntervalInMs")]
```
mikeng13: I think we should name these the same as the actual parameter files

OptimizelySDK.Tests/App.config
msohailhussain: I think this is not for actual HttpConfigManager testing, it's only to check, if the url what we set is readonly through property or not.

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -48,7 +52,61 @@ public static void SetLogger(ILogger logger)
...
             OptimizelyLogger = logger;
         }
+
+        public static Optimizely NewDefaultInstance()
+        {
+            var logger = OptimizelyLogger ?? new DefaultLogger();
+            OptimizelySDKConfigSection OptlySDKConfigSection = null;
+            try
+            {
+                OptlySDKConfigSection = ConfigurationManager.GetSection("optlySDKConfigSection") as OptimizelySDKConfigSection;
```
msohailhussain: Add section name as constant.

OptimizelySDK/Config/OptimizelyConfigHandler.cs
mikeng13: Please remember to address this

PR #206, opened 2020-01-08
feat: Added caching for OptimizelyConfig object

mnoman09:
## Summary
- Updated the OptimizelyConfig implementation to always return cached object. OptimizelyConfig object will only be updated only when new ProjectConfig is received.

## Test plan
Added unit tests


OptimizelySDK/Config/PollingProjectConfigManager.cs
```diff
@@ -145,7 +147,7 @@ public bool SetConfig(ProjectConfig projectConfig)
                 return false;
             
             CurrentProjectConfig = projectConfig;
-
+            CurrentOptimizelyConfig = new OptimizelyConfigService(CurrentProjectConfig).GetOptimizelyConfig();
```
msohailhussain: What if GetOptimizelyConfig or OptimizelyConfigService throws exception? Please add it in try catch and move in separate private method. Also log for exception and reinitializing OptimizelyConfigService

OptimizelySDK/Config/PollingProjectConfigManager.cs
```diff
@@ -29,7 +30,7 @@ namespace OptimizelySDK.Config
     /// Instances of this class, must implement the <see cref="Poll()"/> method
     /// which is responsible for fetching a given ProjectConfig.
     /// </summary>
-    public abstract class PollingProjectConfigManager : ProjectConfigManager, IDisposable
+    public abstract class PollingProjectConfigManager : ProjectConfigManager, IDisposable, IOptimizelyConfigManager
```
msohailhussain: Disposable should come at last.

OptimizelySDK.Tests/OptimizelyConfigTests/OptimizelyConfigTest.cs
```diff
@@ -25,8 +27,72 @@ namespace OptimizelySDK.Tests.OptimizelyConfigTests
...
+        {
+            HttpProjectConfigManager httpManager = new HttpProjectConfigManager.Builder()
+               .WithSdkKey("QBw9gFM8oTn7ogY9ANCC1z")
+               .WithLogger(LoggerMock.Object)
+               .WithPollingInterval(TimeSpan.FromMilliseconds(1000))
+               .WithBlockingTimeoutPeriod(TimeSpan.FromMilliseconds(500))
+               .WithStartByDefault()
+               .Build();
+
+            httpManager.OnReady().Wait(System.Threading.Timeout.Infinite);
```
msohailhussain: Don't need infinite arg, it's by default.

msohailhussain:
Add a testcase where you are polling 2 or more times and everytime assert optimizelyconfig.

OptimizelySDK.Tests/OptimizelyConfigTests/OptimizelyConfigTest.cs
```diff
@@ -93,6 +93,44 @@ public void TestPollingInvalidSdkKeyGetOptimizelyConfig()
...
+            var optimizely = new Optimizely(httpManager);
+
+            var optimizelyConfig = optimizely.GetOptimizelyConfig();
+
+            Assert.NotNull(optimizelyConfig);
+            Assert.NotNull(optimizelyConfig.ExperimentsMap);
+            Assert.NotNull(optimizelyConfig.FeaturesMap);
+            Assert.NotNull(optimizelyConfig.Revision);
+
+            Thread.Sleep(210);
```
msohailhussain: Don't use `Thread.Sleep` add notification and use blocking collection with the timeout to block until specific revision.

PR #211, opened 2020-04-14
feature variable type enum removed

msohailhussain:
## Summary
- FeatureVariableType enum removed.
- Using constant values for Type which are string
- `Type` variable  is now `string` type in `FeatureVariable`.

## Test plan
- Added unit tests which will make sure that additional types shouldn't have any breaking change.
- E2E should pass without modifying any code.

## Issues
- FeatureVariableType and GetFeatureVariableType were public, users who are using these undocumented type may have issues if they upgrade SDK.

### DIFF
--- a/OptimizelySDK.Tests/OptimizelyTest.cs
+++ b/OptimizelySDK.Tests/OptimizelyTest.cs
@@ -1158,7 +1158,7 @@ namespace OptimizelySDK.Tests
             var variableKeyFalse = "varFalse";
             var variableKeyNonBoolean = "varNonBoolean";
             var variableKeyNull = "varNull";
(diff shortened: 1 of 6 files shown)


OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1158,7 +1158,7 @@ public void TestGetFeatureVariableBooleanReturnsCorrectValue()
             var variableKeyFalse = "varFalse";
             var variableKeyNonBoolean = "varNonBoolean";
             var variableKeyNull = "varNull";
-            var featureVariableType = FeatureVariable.VariableType.BOOLEAN;
+            var featureVariableType = "boolean";
```
pawels-optimizely, line 1161: can we use our newly defined types in tests ?

PR #216, opened 2020-05-07
feat: Added GetFeatureVariableJSON.

mnoman09:
## Summary
- Added GetFeatureVariableJson api in optimizely

## Test plan
- Added Unit tests.


OptimizelySDK/Entity/FeatureFlag.cs
```diff
@@ -34,6 +34,13 @@ public List<FeatureVariable> Variables
             }
             set
             {
+                foreach (FeatureVariable variable in value)
```
msohailhussain, line 37: This is not the good place. I will recommend to add it in `FeatureVariable` class. Use it like this. The major issue is, if `FeatureVariable` class is used outside or by someone, will get different result.
```
private string _subType;
        public string SubType { get {
                return _subType;
            }
            set {
                _subType = value;
            }
        }
        private string _type;
        public string Type {
            get {
                 if (_type == STRING_TYPE && _subType == JSON_TYPE) {
                    return JSON_TYPE;
                }
                return _type;
            }
            set {
                _type = value;
            }
         }
```

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1245,6 +1247,25 @@ public void TestGetFeatureVariableStringFRCulture()
...
+        public void TestGetFeatureVariableJSONFRCulture()
+        {
+            SetCulture("en-US");
+            var fallbackConfigManager = new FallbackProjectConfigManager(Config);
+
+            var optimizely = new Optimizely(fallbackConfigManager);
+
+            var optimizelyJsonValue = optimizely.GetFeatureVariableJSON("string_single_variable_feature", "json_var", "testUser1");
+
+            Assert.AreEqual(optimizelyJsonValue.GetValue<long>("int_var"), 1);
```
msohailhussain, line 1260: I will suggest, use TestData.CompareResult. 
```var expectedPayload = new Dictionary<int, object>() {
  {"int_var", 1},
  {"boolean_key", false}
};
var actualValue = optimizelyJsonValue.ToDictionary();
// Compare them.
```
The reason I am telling you, it will give false impression if GetValue is not working as expected.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1316,6 +1337,32 @@ public void TestGetFeatureVariableStringReturnsCorrectValue()
...
+        {
+            var featureKey = "featureKey";
+            var variableKeyString = "varJSONString1";
+            var variableKeyIntString = "varJSONString2";
+            var variableKeyDouble = "varJSONDouble";
+            var variableKeyNull = "varNull";
+            var featureVariableType = "json";
+
+            OptimizelyMock.Setup(om => om.GetFeatureVariableValueForType<OptimizelyJson>(It.IsAny<string>(), variableKeyString, It.IsAny<string>(),
+                It.IsAny<UserAttributes>(), featureVariableType)).Returns(new OptimizelyJson("{\"string\": \"Test String\"}", ErrorHandlerMock.Object, LoggerMock.Object));
```
msohailhussain, line 1351: take this in a dictionary and use the same approach. comparedata.
`{\"string\": \"Test String\"}"`

msohailhussain: Please do it same thing for all.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1624,7 +1704,7 @@ public void TestUnsupportedVariableType()
             Assert.IsNull(featureVariableStringRandomType);
 
             var featureVariableStringJsonType = Optimizely.GetFeatureVariableString("unsupported_variabletype", "string_json_key", TestUserId);
-            Assert.AreEqual(featureVariableStringJsonType, "{\"myvalue\": \"jsonValue\"}");            
+            Assert.IsNull(featureVariableStringJsonType);
```
msohailhussain, line 1707: Why value is changed to `IsNULL`

mnoman09: because the subtype of that variable is json and we are changing type from string to json upon initialization. so it will not match and will return null as type of variable will not match

OptimizelySDK.Tests/TestData.json
```diff
@@ -589,6 +593,13 @@
 					"key": "string_variable",
 					"type": "string",
 					"defaultValue": "wingardium leviosa"
+				},
```
pawels-optimizely, line 596: since C# had a problem before with json type, please add
```{
					"defaultValue": "{\"int_var\": 5212, \"boolean_key\": true}",
					"type": "json",
					"id": "1",
					"key": "true_json_var"
				}
``` 
and test

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1444,6 +1545,38 @@ public void TestGetFeatureVariableBooleanReturnsRightValueWhenUserBuckedIntoRoll
             LoggerMock.Verify(l => l.Log(LogLevel.INFO, $@"Returning variable value ""true"" for variation ""{variation.Key}"" of feature flag ""{featureKey}""."));
         }
 
+        [Test]
+        public void TestGetFeatureVariableJSONReturnsRightValueWhenUserBuckedIntoRolloutAndVariationIsToggleOn()
```
msohailhussain, line 1549: TestGetFeatureVariableJSONReturnsRightValueWhenUserBucke`t`IntoRolloutAndVariationIsToggleOn

OptimizelySDK.Tests/TestData.json
mnoman09: Added the test you asked for. So can you please review if this is what you asked for?  [link](https://github.com/optimizely/csharp-sdk/pull/216/commits/4758fbdc3ca38db6481f5e44437fec8f46062de4)

PR #217, opened 2020-05-13
feat: Added GetAllFeatureVariables API and NotificationListener support 

mnoman09:
## Summary
- Added getAllFeatureVariables Api which Get the values of all variables in the feature 
- Decision Notification of type all-feature-variable is getting sent in getAllFeatureVariables API. 
- Sending map of optimizely json in decision notification of type all-feature-variable

## Test Plan
- FSC should pass with FEATURE_JSON tag.
- Updated unit tests accordingly.


OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1789,6 +1789,9 @@ public void TestUnsupportedVariableType()
             var featureVariableStringRandomType = Optimizely.GetFeatureVariableString("", "any_key", TestUserId);
             Assert.IsNull(featureVariableStringRandomType);
 
+            var featureVariableStringJsonType = Optimizely.GetFeatureVariableString("unsupported_variabletype", "string_json_key", TestUserId);
```
msohailhussain, line 1792: what's the point to add test here?

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
             NotificationCallbackMock.Verify(nc => nc.TestDecisionCallback(DecisionNotificationTypes.FEATURE_VARIABLE, TestUserId, new UserAttributes(), It.Is<Dictionary<string, object>>(info => TestData.CompareObjects(info, decisionInfo))), Times.Once);
         }
 
+        [Test]
+        public void TestGetAllFeatureVariablesSendsNotificationWhenUserBuckedIntoRolloutAndVariationIsToggleOn()
```
msohailhussain, line 3198: TestGetAllFeatureVariablesSendsNotificationWhenUserBucke`t`IntoRolloutAndVariationIsToggleOn

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+               { "location", "San Francisco" }
+            };
+
+            DecisionServiceMock.Setup(ds => ds.GetVariationForFeature(featureFlag, TestUserId, Config, userAttributes)).Returns(decision);
+            NotificationCallbackMock.Setup(nc => nc.TestDecisionCallback(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<UserAttributes>(),
+                It.IsAny<Dictionary<string, object>>()));
+
+            var optly = Helper.CreatePrivateOptimizely();
+
+            optly.SetFieldOrProperty("ProjectConfigManager", ConfigManager);
```
msohailhussain, line 3227: I don't know why we are using `reflection` to instantiate `optimizely` instance. please explain

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+               { "device_type", "iPhone" },
+               { "company", "Optimizely" },
+               { "location", "San Francisco" }
+            };
+            var optly = Helper.CreatePrivateOptimizely();
+
+            var result = (OptimizelyJson)optly.Invoke("GetAllFeatureVariables", null, TestUserId, userAttributes);
+            Assert.Null(result);
+
+            LoggerMock.Verify(log => log.Log(LogLevel.WARN, "The featureKey parameter must be nonnull."), Times.Once);
```
msohailhussain, line 3265: `nonnull` or `not-null` @pawels-optimizely need suggestion.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+        }
+
+        [Test]
+        public void TestGetAllFeatureVariablesNullUserID()
```
msohailhussain, line 3269: I think add all `null` scenarios in a single method. Secondly don't use createPrivateOptimizely, it is being used if we need to get or set any private value.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+            var userAttributes = new UserAttributes
+            {
+               { "device_type", "iPhone" },
+               { "company", "Optimizely" },
+               { "location", "San Francisco" }
+            };
+
+            var featureFlag = Config.GetFeatureFlagFromKey(featureKey);
+
+            var decision = new FeatureDecision(experiment, null, FeatureDecision.DECISION_SOURCE_ROLLOUT);
```
msohailhussain, line 3336: Why variation is null.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+
+            DecisionServiceMock.Setup(ds => ds.GetVariationForFeature(featureFlag, TestUserId, Config, userAttributes)).Returns(decision);
+
+            var optly = Helper.CreatePrivateOptimizely();
```
msohailhussain, line 3340: revise it.

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -3191,8 +3194,185 @@ public void TestGetFeatureVariableDoubleSendsNotificationWhenUserNotBuckedIntoBo
...
+
+            var result = (OptimizelyJson)optly.Invoke("GetAllFeatureVariables", featureKey, TestUserId, userAttributes);
+            Assert.NotNull(result);
+
+            LoggerMock.Verify(log => log.Log(LogLevel.INFO, "User \"" + TestUserId + "\" was not bucketed into any variation for feature flag \"" + featureKey + "\". " +
+                        "The default values are being returned."), Times.Once);
+        }
+
+        [Test]
+        public void TestGetAllFeatureVariablesValid()
```
msohailhussain, line 3352: What does it mean valid?

OptimizelySDK.Tests/OptimizelyTest.cs
mnoman09: It was an old test which was testing that getFeatureVariableString will return string value  even when given json as subtype but this is no longer the case here. So now if the sub type is JSON and we will call getFeatureVariableString to get that variable, it will return null because type will not be matched.

OptimizelySDK.Tests/OptimizelyTest.cs
mnoman09: this is to set decisionservice object to decisionserviceMock object.

OptimizelySDK.Tests/OptimizelyTest.cs
mnoman09: To verify info log that ""User \"" + TestUserId + "\" was not bucketed into any variation for feature flag \"" + featureKey + "\". The default values are being returned.""

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -1788,7 +1788,7 @@ public void TestUnsupportedVariableType()
         {
             var featureVariableStringRandomType = Optimizely.GetFeatureVariableString("", "any_key", TestUserId);
             Assert.IsNull(featureVariableStringRandomType);
-
+
```
msohailhussain, line 1791: remove spacing.

OptimizelySDK.Tests/OptimizelyTest.cs
msohailhussain: got it.

OptimizelySDK.Tests/OptimizelyTest.cs
pawels-optimizely: cannot be null

PR #223, opened 2020-06-24
fix: http project config manager unit tests revised.

msohailhussain:
## Summary
Unit tests were using real datafiles which may cause failure due to network issues. I have mocked all SendAsync to provide hardcoded datafiles when called from HttpClient.

## Test plan
Unit test must pass.


OptimizelySDK.Tests/ConfigTest/HttpProjectConfigManagerTest.cs
```diff
@@ -50,16 +50,21 @@ public void Setup()
...
         {
+            var t = MockSendAsync(TestData.Datafile);
             HttpProjectConfigManager httpManager = new HttpProjectConfigManager.Builder()
                 .WithUrl("https://cdn.optimizely.com/datafiles/QBw9gFM8oTn7ogY9ANCC1z.json")
                 .WithLogger(LoggerMock.Object)
                 .WithPollingInterval(TimeSpan.FromMilliseconds(1000))
                 .WithBlockingTimeoutPeriod(TimeSpan.FromMilliseconds(500))
                 .WithStartByDefault()
                 .Build();
+            t.Wait(1000);
```
mikeng13, line 61: Do we need to wait here if we are mocking the request? Can we continue using `httpManager.OnReady()`? It seems like we only need to wait on something because the operation is async

OptimizelySDK.Tests/ConfigTest/HttpProjectConfigManagerTest.cs
```diff
@@ -81,13 +88,19 @@ public void TestHttpConfigManagerRetreiveProjectConfigGivenEmptyFormatUseDefault
                  .WithStartByDefault()
                  .Build();
 
-            httpManager.OnReady().Wait(System.Threading.Timeout.Infinite);
-            Assert.NotNull(httpManager.GetConfig());
+            t.Wait(1000);
```
mikeng13, line 91: Same as above

OptimizelySDK.Tests/ConfigTest/HttpProjectConfigManagerTest.cs
```diff
@@ -129,88 +157,111 @@ public void TestOnReadyPromiseResolvedImmediatelyWhenDatafileIsProvided()
...
             httpManager.OnReady().Wait();
-            stopwatch.Stop();
 
-            Assert.True(stopwatch.Elapsed.Seconds == 0);
-            Assert.NotNull(httpManager.GetConfig());
+            Assert.AreEqual("15", httpManager.GetConfig().Revision);
+
+            // loaded datafile from config manager after a second.
+            t.Wait();
+            Task.Delay(200).Wait();
```
mikeng13, line 166: Heh i am not comfortable with us introducing all these arbitrary waits. Is there something else we can use as a trigger? Like maybe wait for a config-update notif?

msohailhussain:
Added more proper comments and reason for wait usage.

PR #235, opened 2020-07-13
fix: When rollout has no rule, should return null

msohailhussain:
## Summary
- When there is no rollout rule, it should getVariationForFeatureRollout should return false.

## Test plan
- FSC and all Unit tests must pass.
- Added one unit test.


OptimizelySDK.Tests/EmptyRolloutRule.json
```diff
@@ -0,0 +1,44 @@
...
+      "name": "Optimizely-Generated Audience for Backwards Compatibility"
+    }
+  ],
+  "groups": [],
+  "attributes": [],
+  "botFiltering": false,
+  "accountId": "8272261422",
+  "events": [],
+  "revision": "2"
+}
```
msohailhussain, line 44: need to add EOL.

PR #241, opened 2020-09-02
fix: error handler not raise exception while using optimizelyfactory

msohailhussain:
## Summary
- NewDefaultInstance shouldn't cause optimizely instance to raise exception while handling error, by default all DefaultErrorHandler shouldn't raise exception.
- Passed logger in default error handler and made raise exception to false.

## Test plan
- Added unit test. 
- FSC with empty datafile scenario must pass.
   https://travis-ci.com/github/optimizely/fullstack-sdk-compatibility-suite/builds/182487252


mikeng13 (approved):
lgtm

PR #242, opened 2020-09-03
fix: Semantic version fix and multiple dash signs check added

mnoman09:
## Summary
- Multiple fixed for sem-versioning
- Compare build version and pre-release fix
- Multiple `-` is now handled.

## Test plan
All FSC tests must pass
Added unit tests

### DIFF
--- a/OptimizelySDK/AudienceConditions/SemanticVersion.cs
+++ b/OptimizelySDK/AudienceConditions/SemanticVersion.cs
@@ -44,7 +44,17 @@ namespace OptimizelySDK.AudienceConditions
         /// <returns>True if Semantic version contains '-', else false</returns>
         public static bool IsPreRelease(this string semanticVersion)
         {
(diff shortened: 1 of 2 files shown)


OptimizelySDK/AudienceConditions/SemanticVersion.cs
```diff
@@ -54,7 +64,17 @@ public static bool IsPreRelease(this string semanticVersion)
...
+            var preReleaseIndex = semanticVersion.IndexOf(PreReleaseSeparator.ToString());
+            if (preReleaseIndex < 0)
+            {
+                return buildIndex > 0;
+            }
+            else if (buildIndex < 0)
+            {
+                return false;
+            }
+            return buildIndex < preReleaseIndex;
```
aliabbasrizvi, line 77: A unified `IsPrereleaseOrBuild` method may be slightly more performant. I'd recommend going down that path.

thomaszurkan-optimizely (approved):
LGTM

PR #253, opened 2020-12-08
Feat: Add thread-safe UserAttributes in read/write operations.

mnoman09:
## Summary
- Copying UserAttributes to make it thread safe.
- Made All fields in OptimizelyUserContext private and added getter methods

## Test plan
Updated all Unit tests


OptimizelySDK/OptimizelyUserContext.cs
```diff
@@ -46,6 +46,21 @@ public OptimizelyUserContext(Optimizely optimizely, string userId, UserAttribute
...
+        }
+
+        public string GetUserId()
+        {
+            return UserId;
+        }
+
+        public UserAttributes GetAttributes()
+        {
+            return new UserAttributes(Attributes);
```
jaeopt, line 61: Don't we need a mutex protection for this copy with the same lock for setAttribute?

jaeopt (changes requested):
Missing a mutex lock while making a copy. Other than that, all looks good!

OptimizelySDK/OptimizelyUserContext.cs
msohailhussain, line 61: yes, mutex needed.

msohailhussain (approved):
lgtm

jaeopt (approved):
LGTM

PR #254, opened 2020-12-11
Feat: Added unit tests for reasons addInfo logs 

mnoman09:
## Summary
- Added few unit tests to test that DecisionReasons contains the info logs upon setting decideOption as INCLUDE_REASONS

## Test plan
It should pass with FSC and all unit tests should pass


jaeopt (approved):
LGTM

PR #256, opened 2021-01-12
feat(decide): clone user-context before calling optimizely decide

mnoman09:
Clone the current user-context and pass it to optimizely.decide(), decide all and in decideForKeys
- user context can be changed while processing the decide request.
- with this fix, decide can return the copy of the original user-context as a part of the decision

### DIFF
--- a/OptimizelySDK/OptimizelyUserContext.cs
+++ b/OptimizelySDK/OptimizelyUserContext.cs
@@ -46,7 +46,7 @@ namespace OptimizelySDK
             UserId = userId;
         }
 
-        private OptimizelyUserContext Clone() => new OptimizelyUserContext(Optimizely, UserId, Attributes, ErrorHandler, Logger);
+        private OptimizelyUserContext Copy() => new OptimizelyUserContext(Optimizely, UserId, GetAttributes(), ErrorHandler, Logger);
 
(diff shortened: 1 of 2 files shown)


OptimizelySDK/OptimizelyUserContext.cs
```diff
@@ -46,6 +46,8 @@ public OptimizelyUserContext(Optimizely optimizely, string userId, UserAttribute
             UserId = userId;
         }
 
+        private OptimizelyUserContext Clone() => new OptimizelyUserContext(Optimizely, UserId, Attributes, ErrorHandler, Logger);
```
jaeopt, line 49: change "Attributes" to "getAttributes()", which should be mutex protected while copying.

jaeopt (changes requested):
Looks good. A change suggested.

OptimizelySDK/OptimizelyUserContext.cs
msohailhussain: Also rename it to Copy.

msohailhussain (approved):
lgtm. please address jae's comments.

jaeopt (approved):
LGTM

PR #257, opened 2021-01-27
Fix: SendImpressionEvent will return false when event is not sent

mnoman09:
## Summary
- SendImpressionEvent will return false when event is not sent and return true when sent successfully
- In case of userEvent is null it will return false


jaeopt (changes requested):
Looks good. Can we add unit tests to catch the error?

jaeopt (approved):
LGTM

PR #258, opened 2021-02-01
Feat: Added unit test to verify in which SendFlagDecision is true and false

mnoman09:
## Summary
- Added unit tests to verify when source type is rollout and SendFlagDecisions is false then event should not dispatch and decisionEventDispatched should be false.
- Added unit tests to verify when source type is rollout and SendFlagDecisions is true then event should dispatch and decisionEventDispatched should be true.

### DIFF
--- a/OptimizelySDK.Tests/OptimizelyTest.cs
+++ b/OptimizelySDK.Tests/OptimizelyTest.cs
@@ -241,12 +241,90 @@ namespace OptimizelySDK.Tests
         }
 
         [Test]
+        public void TestDecisionNotificationSentWhenSendFlagDecisionsFalseAndFeature()
+        {
+            var featureKey = "boolean_feature";
+            var variables = Optimizely.GetAllFeatureVariables(featureKey, TestUserId);
+            var userAttributes = new UserAttributes
+            {
+                { "device_type", "iPhone" },
+                { "location", "San Francisco" }
+            };
+            Config.SendFlagDecisions = false;
+            var fallbackConfigManager = new FallbackProjectConfigManager(Config);
+            var optimizely = new Optimizely(fallbackConfigManager,
+                NotificationCenter,
+                EventDispatcherMock.Object,
+                LoggerMock.Object,
+                ErrorHandlerMock.Object,
+                null,
+                new ForwardingEventProcessor(EventDispatcherMock.Object, NotificationCenter, LoggerMock.Object, ErrorHandlerMock.Object),
+                null);
+
+            // Mocking objects.
+            NotificationCallbackMock.Setup(nc => nc.TestDecisionCallback(It.IsAny<string>(), It.IsAny<string>(),
+               It.IsAny<UserAttributes>(), It.IsAny<Dictionary<string, object>>()));
+            optimizely.NotificationCenter.AddNotification(NotificationCenter.NotificationType.Decision, NotificationCallbackMock.Object.TestDecisionCallback);
+
+            var optimizelyUserContext = optimizely.CreateUserContext(TestUserId, userAttributes);
+            optimizelyUserContext.Decide(featureKey);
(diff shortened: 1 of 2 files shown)


OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -240,6 +240,94 @@ public void TestCreateUserContextMultipleAttribute()
             Assert.AreEqual(attribute2, optlyUserContext2.GetAttributes());
         }
 
+        [Test]
+        public void TestDecisionNotificationNotSentWhenSendFlagDecisionsFalseAndRollout()
+        {
+            var featureKey = "boolean_single_variable_feature";
+            var featureFlag = Config.GetFeatureFlagFromKey(featureKey);
+            var variables = Optimizely.GetAllFeatureVariables(featureKey, TestUserId);
+            var reasons = new OptimizelyDecideOption[0];
```
jaeopt, line 249: Remove this line

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -240,6 +240,94 @@ public void TestCreateUserContextMultipleAttribute()
...
+            }))), Times.Once);
+        }
+
+        [Test]
+        public void TestDecisionNotificationSentWhenSendFlagDecisionsTrueAndRollout()
+        {
+            var featureKey = "boolean_single_variable_feature";
+            var featureFlag = Config.GetFeatureFlagFromKey(featureKey);
+            var variables = Optimizely.GetAllFeatureVariables(featureKey, TestUserId);
+            var reasons = new OptimizelyDecideOption[0];
```
jaeopt, line 293: Remove

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -240,6 +240,94 @@ public void TestCreateUserContextMultipleAttribute()
...
+            optimizelyUserContext.Decide(featureKey);
+            NotificationCallbackMock.Verify(nc => nc.TestDecisionCallback(DecisionNotificationTypes.FLAG, TestUserId, userAttributes, It.Is<Dictionary<string, object>>(info => TestData.CompareObjects(info, new Dictionary<string, object> {
+                { "flagKey", featureKey },
+                { "enabled", true },
+                { "variables", variables.ToDictionary() },
+                { "variationKey", variation.Key },
+                { "ruleKey", ruleKey },
+                { "reasons", reasons },
+                { "decisionEventDispatched", false }
+            }))), Times.Once);
```
jaeopt, line 284: Can you validation that event is actually not sent out?

OptimizelySDK.Tests/OptimizelyTest.cs
```diff
@@ -240,6 +240,94 @@ public void TestCreateUserContextMultipleAttribute()
...
+            NotificationCallbackMock.Verify(nc => nc.TestDecisionCallback(DecisionNotificationTypes.FLAG, TestUserId, userAttributes, It.Is<Dictionary<string, object>>(info => TestData.CompareObjects(info, new Dictionary<string, object> {
+                { "flagKey", featureKey },
+                { "enabled", true },
+                { "variables", variables.ToDictionary() },
+                { "variationKey", variation.Key },
+                { "ruleKey", ruleKey },
+                { "reasons", reasons },
+                { "decisionEventDispatched", true }
+            }))), Times.Once);
+        }
```
jaeopt, line 329: Can you validation that event is actually sent out and event contents as well?

jaeopt: Also, can we add the same couple of tests for "featureTest" (not rollout) as well?

jaeopt (changes requested):
Tests look good. 
Suggested to add the same tests for featureTest as well.

jaeopt (approved):
LGTM

PR #260, opened 2021-03-16
fix: DefaultErrorHandler shouldn't raise exception

msohailhussain:
## Summary
- When no ErrorHandler is provided in the HttpProjectConfigManager, it provides DefaultErrorHandler that raise exception without providing any argument. 
- Fixed by explicitly setting throwException=false in the constructor.

## Test plan
- Added a unit test.
- All FSC should be passed.


jaeopt (approved):
LGTM

PR #262, opened 2021-03-25
fix: nuget added configuration manager dependency for .netstandard 2.0

msohailhussain:
## Summary
- System.Configuration.ConfigurationManager dependency was missing in nuget package.
- licenseUrl is deprecated, updated to license element.
- IconUrl is deprecated, updated to Icon

## Test plan
- All unit tests must pass
- Manual testing, generated nuget package and tested 
- FSC must pass.


jaeopt (approved):
LGTM

PR #264, opened 2021-05-06
Feat: Added set polling interval and set blocking timeout period method in OptimizelyFactory

mnoman09:
## Summary
- Polling interval and blocking timeout period of configManager can be now set using optimizelyFactory and the provided intervals will be given priority over config file values. 

## Test plan
All Unit tests and FSC tests should pass.

## Issues
- [OASIS-7690](https://optimizely.atlassian.net/browse/OASIS-7690)

### DIFF
--- a/OptimizelySDK.Tests/OptimizelyFactoryTest.cs
+++ b/OptimizelySDK.Tests/OptimizelyFactoryTest.cs
@@ -63,54 +63,6 @@ namespace OptimizelySDK.Tests
         }
 
         [Test]
-        public void TestOptimizelyInstanceUsingConfigFileWithPollingInterval()
-        {
-            OptimizelyFactory.SetPollingInterval(TimeSpan.FromMilliseconds(2023));
-            var optimizely = OptimizelyFactory.NewDefaultInstance();
-            // Check values are loaded from app.config or not.
-            var projectConfigManager = optimizely.ProjectConfigManager as HttpProjectConfigManager;
-            Assert.NotNull(projectConfigManager);
-
-            var actualConfigManagerProps = new ProjectConfigManagerProps(projectConfigManager);
-            var expectedConfigManagerProps = new ProjectConfigManagerProps
-            {
-                Url = "www.testurl.com",
-                LastModified = "",
-                AutoUpdate = true,
-                DatafileAccessToken = "testingtoken123",
-                BlockingTimeout = TimeSpan.FromSeconds(10),
-                PollingInterval = TimeSpan.FromMilliseconds(2023)
-            };
-
-            Assert.AreEqual(actualConfigManagerProps, expectedConfigManagerProps);
-            optimizely.Dispose();
-        }
-
-        [Test]
-        public void TestOptimizelyInstanceUsingConfigFileWithBlockingTimeOutInterval()
-        {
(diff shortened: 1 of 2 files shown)


OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -76,13 +88,12 @@ public static Optimizely NewDefaultInstance()
...
             var builder = new HttpProjectConfigManager.Builder();
             var notificationCenter = new NotificationCenter();
-            
             var configManager = builder
                 .WithSdkKey(httpProjectConfigElement.SDKKey)
                 .WithUrl(httpProjectConfigElement.Url)
                 .WithFormat(httpProjectConfigElement.Format)
-                .WithPollingInterval(TimeSpan.FromMilliseconds(httpProjectConfigElement.PollingInterval))
-                .WithBlockingTimeoutPeriod(TimeSpan.FromMilliseconds(httpProjectConfigElement.BlockingTimeOutPeriod))
+                .WithPollingInterval(PollingInterval == TimeSpan.Zero ? TimeSpan.FromMilliseconds(httpProjectConfigElement.PollingInterval) : PollingInterval)
```
msohailhussain, line 95: PLease update header.

OptimizelySDK.Tests/OptimizelyFactoryTest.cs
```diff
@@ -102,15 +101,40 @@ public void TestProjectConfigManagerWithDatafileAccessToken()
...
+                BlockingTimeout = TimeSpan.FromSeconds(30),
+                PollingInterval = TimeSpan.FromMilliseconds(2023)
             };
 
             Assert.AreEqual(actualConfigManagerProps, expectedConfigManagerProps);
 
             optimizely.Dispose();
         }
 
+        [Test]
```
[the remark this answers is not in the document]
mnoman09: these tests already show that configElement values will be discarded once we set the polling interval using optimizelyFactory.
Value given in ConfigElement is 2 second, where as we can see that at line 130 and 116 after setting polling Interval using OptlyFactory all other values remained same only polling and blocking interval are changed.

msohailhussain (approved):
lgtm

OptimizelySDK/OptimizelyFactory.cs
```diff
@@ -128,6 +139,8 @@ public static Optimizely NewDefaultInstance(string sdkKey, string fallback, stri
                 .WithSdkKey(sdkKey)
                 .WithDatafile(fallback)
                 .WithLogger(logger)
+                .WithPollingInterval(PollingInterval)
+                .WithBlockingTimeoutPeriod(BlockingTimeOutPeriod)
```
jaeopt, line 143: I'm confused - why don't have the config reading unlike the above: 
                **.WithPollingInterval(PollingInterval == TimeSpan.Zero ? TimeSpan.FromMilliseconds(httpProjectConfigElement.PollingInterval) : PollingInterval)**
This can have different configuration paths depending on factory method parameters. It can be confusing unless we document very clearly about them.

jaeopt (changes requested):
It looks good. One concern is we have 3 similar factory methods, which can have totally different config data paths (see comments) for timeout values. Can we merge and make them consistent?

jaeopt (approved):
LGTM

PR #279, opened 2021-09-01
fix: experimentRules should have same order as experimentIds

msohailhussain:
## Summary
- ExperimentRules sequence critically important when evaluating for decision. it should follow experimentIds under featureFlag object sequence.

## Test plan
- We have test cases in FSC, must pass for unsorted scenarios.

### DIFF
--- a/OptimizelySDK/OptlyConfig/OptimizelyConfigService.cs
+++ b/OptimizelySDK/OptlyConfig/OptimizelyConfigService.cs
@@ -227,12 +227,9 @@ namespace OptimizelySDK.OptlyConfig
             foreach (var featureFlag in projectConfig.FeatureFlags)   
             {
 
-                var featureExperimentMap = new Dictionary<string, OptimizelyExperiment>();
-                foreach (var experimentId in featureFlag.ExperimentIds)
-                {
-                    var experimentDictionary = experimentsMapById[experimentId];
-                    featureExperimentMap.Add(experimentDictionary.Key, experimentDictionary);
-                }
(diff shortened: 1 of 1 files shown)


OptimizelySDK/OptlyConfig/OptimizelyConfigService.cs
```diff
@@ -227,8 +227,8 @@ private OptimizelyAttribute[] GetAttributes(ProjectConfig projectConfig)
             foreach (var featureFlag in projectConfig.FeatureFlags)   
             {
 
-                var featureExperimentMap = experimentsMapById.Where(expMap => featureFlag.ExperimentIds.Contains(expMap.Key))
-                    .ToDictionary(k => k.Value.Key, v => v.Value);
+                var featureExperimentMap = featureFlag.ExperimentIds.Select(experimentId => experimentsMapById[experimentId])
+                    .ToDictionary(experiment => experiment.Key, experiment => experiment);
```
jaeopt, line 231: It looks like losing order here - it picks up experimentIds as an array, but sets to map immediately.

jaeopt (changes requested):
Can you confirm the exp ordering?

OptimizelySDK/OptlyConfig/OptimizelyConfigService.cs
msohailhussain: yes i have run multiple times with unsorted scenario and works fine.

jaeopt (approved):
LGTM

dustin-sier (approved):
LGTM, really appreciate that change for both the feature map and the experiment rules, much easier to work forward from.

PR #280, opened 2021-09-06
Create virtual methods on user context

daniel-pons:
The following versions of Optimizely C# SDK would use the methods defined on the class _OptimizelyUserContext_ to perform some processes, E.G: Decide, TrackEvent, DecideAll, etc...

Then, it is required on the vast majority of unit testing frameworks, for mocking this object to be able to override such members.

This can be accomplished, at least on C#, usually in two ways:
1. Declaring an _interface_ (with the methods to be overriden) and having the class implementing it.
2. Declaring the members as _virtual_, would allow also allow the testing framework to override them.

This PR uses the second approach, it makes the methods on _OptimizelyUserContext_ virtual.

Also fixes a problem with development environments with a culture different from English.

### DIFF
--- a/OptimizelySDK.Tests/OptimizelySDK.Tests.csproj
+++ b/OptimizelySDK.Tests/OptimizelySDK.Tests.csproj
@@ -97,6 +97,7 @@
     <Compile Include="TestBucketer.cs" />
     <Compile Include="BucketerTest.cs" />
     <Compile Include="ProjectConfigTest.cs" />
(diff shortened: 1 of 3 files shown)


msohailhussain:
can you please update it with master.

daniel-pons:
> can you please update it with master.

My bad, done!

msohailhussain (approved):
lgtm

daniel-pons:
Many thanks!

PR #287, opened 2021-11-30
Refact: Refactored Forced decision

mnoman09:
## Summary
- Refactored code and fixed some minor issues. 
- Introduced ForcedDecisionStore class that will be used for storing forced decisions.
- Removed DecisionKey, instead introduced map of map in forceddecisionstore class. 
- GetFlagVariationKey removed from optimizely method, single source of truth for all mappings is DatafileProjectConfig class
- Added config in the FindValidatedForcedDecision since no access of ProjectConfig directly from OptimizelyUserContext.
- Refactored DecisionService, not needed for GetVariationFromExperimentRules or GetVariationFromDeliveryRules etc ... 
- No need of Separate mapping method in DatafileProjectConfig, tried to use existing maps to prepare FlagVariationsMap.
- Documentation corrected.

## Test plan
All FSC and unit tests should pass

## Issues
- I still see a little room of refactoring in DecisionService but for now this is fine.

### DIFF
--- a/OptimizelySDK/ForcedDecisionsStore.cs
+++ b/OptimizelySDK/ForcedDecisionsStore.cs
@@ -1,0 +1,80 @@
+﻿/*
+ * Copyright 2021, Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Collections.Generic;
+
+namespace OptimizelySDK
+{
+    /// <summary>
+    /// ForcedDecisionsStore defines helper methods that are used for optimizelyDecisionContext object manipulation.
+    /// </summary>
+    public class ForcedDecisionsStore
+    {
+        public const string OPTI_NULL_RULE_KEY = "$opt-null-rule-key";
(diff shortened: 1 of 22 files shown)


OptimizelySDK/OptimizelyUserContext.cs
```diff
@@ -41,29 +41,24 @@ public class OptimizelyUserContext
...
-            ErrorHandler = errorHandler;
-            Logger = logger;
-            Optimizely = optimizely;
-            Attributes = userAttributes ?? new UserAttributes();
-            ForcedDecisionsMap = new Dictionary<string, OptimizelyForcedDecision>();
-            UserId = userId;
         }
 
-        public OptimizelyUserContext(Optimizely optimizely, string userId, UserAttributes userAttributes, Dictionary<string, OptimizelyForcedDecision> forcedDecisions, IErrorHandler errorHandler, ILogger logger)
+        public OptimizelyUserContext(Optimizely optimizely, string userId, UserAttributes userAttributes, ForcedDecisionsStore forcedDecisionsStore, IErrorHandler errorHandler, ILogger logger)
```
msohailhussain, line 51: Do we really need this constructor method? @jaeopt  a question for you.

OptimizelySDK/ForcedDecisionsStore.cs
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+using System.Collections.Generic;
+
+namespace OptimizelySDK
+{
+    /// <summary>
+    /// ForcedDecisionsStore defines helper methods that are used for optimizelyDecisionContext object manipulation.
+    /// </summary>
+    public class ForcedDecisionsStore
```
msohailhussain, line 24: Unit tests are already covered for this class as transitive dependency. But I will appreciate @mnoman09 to write unit tests.

jaeopt (changes requested):
Refactor looks good! Some changes suggested.

dustin-sier:
LGTM, appreciate the changes

jaeopt (approved):
LGTM

PR #300, opened 2022-05-23
Added linter for csharp

mnoman09:
## Summary
- Added superLinter to verify markdown files and .cs files using editorConfig file 
- Removed markdown lint stage as it is no longer needed 
- Linter will not verify existing issues as VALIDATE_ALL_CODEBASE is set to false.
- editorconfig file contains configuration for IDE and gives suggestions and warning related to linting to make sure user follows the rules.


.github/workflows/csharp.yml
```diff
@@ -1,28 +1,31 @@
...
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          # Full git history is needed to get a proper list of changed files
+          fetch-depth: 0
+      - name: Lint Code Base
+        uses: github/super-linter@v4
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: master
```
msohailhussain, line 24: does this mean it only checks for master branch?

.github/workflows/csharp.yml
```diff
@@ -1,28 +1,31 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
msohailhussain, line 25: Why github token is needed?

.github/workflows/csharp.yml
mnoman09: yes It will compare and check linting of only the files which were modified and compares with master.

msohailhussain (approved):
looks cool.

PR #315, opened 2022-10-10
feat: ODP REST API for Sending ODP Events

mikechu-optimizely:
## Summary
Adding module to provide an internal service for sending events to ODP's REST API.

## Test plan

- Added new unit tests to .spec
- All Existing Unit tests pass.
- All existing full-stack compatibility tests pass.

## Issues
- FSSDK-8449


mikechu-optimizely:
@msohailhussain Let me know if you want to pair up over this review.

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -94,15 +135,101 @@ public GraphQLManager(IErrorHandler errorHandler = null, ILogger logger = null,
...
+        /// <returns>GraphQL payload</returns>
+        private static string ToGraphQLJson(string userKey, string userValue,
+            IEnumerable segmentsToCheck
+        )
+        {
+            var userValueWithEscapedQuotes = $"\\\"{userValue}\\\"";
+            var segmentsArrayJson =
+                JsonConvert.SerializeObject(segmentsToCheck).Replace("\"", "\\\"");
+
+            var json = new StringBuilder();
```
msohailhussain, line 149: can we revise this approach?
e.g. 
Function should take children. and append the children inside function. "Closing and Opening brackets" may cause issues in this kind of situation.

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -94,15 +135,101 @@ public GraphQLManager(IErrorHandler errorHandler = null, ILogger logger = null,
...
+                response = QuerySegmentsAsync(apiKey, endpoint, query).GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
+            {
+                _errorHandler.HandleError(ex);
+                _logger.Log(LogLevel.ERROR, $"{AUDIENCE_FETCH_FAILURE_MESSAGE} (network error)");
+                return default;
+            }
+
+            var responseStatusCode = (int)response.StatusCode;
```
msohailhussain, line 179: Isn't this property blocking async request?

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -94,15 +135,101 @@ public GraphQLManager(IErrorHandler errorHandler = null, ILogger logger = null,
...
+        /// <param name="endpoint">Fully-qualified ODP GraphQL endpoint URL</param>
+        /// <param name="query">GraphQL query</param>
+        /// <returns>JSON response from ODP</returns>
+        private async Task<HttpResponseMessage> QuerySegmentsAsync(string apiKey, string endpoint,
+            string query
+        )
+        {
+            var request = BuildRequestMessage(apiKey, endpoint, query);
+
+            var response = await _httpClient.SendAsync(request);
```
msohailhussain, line 203: what if you return?
return _httpClient.SendAsync(request);

OptimizelySDK/Odp/RestApiManager.cs
```diff
@@ -0,0 +1,164 @@
...
+                response = SendEventsAsync(apiKey, endpoint, data).GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
+            {
+                _errorHandler.HandleError(ex);
+                _logger.Log(LogLevel.ERROR, $"{EVENT_SENDING_FAILURE_MESSAGE} (network error)");
+                shouldRetry = true;
+            }
+
+            var responseStatusCode = response == null ? 0 : (int)response.StatusCode;
```
msohailhussain, line 104: same ensurestatuscode

OptimizelySDK/Odp/RestApiManager.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (responseStatusCode >= 400)
```
msohailhussain, line 105: I am not sure if this should be outside of catch block.

OptimizelySDK/Odp/RestApiManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <param name="endpoint">Fully-qualified ODP REST API endpoint</param>
+        /// <param name="data">JSON string version of ODP event data</param>
+        /// <returns>HTTP response endpoint</returns>
+        private async Task<HttpResponseMessage> SendEventsAsync(string apiKey, string endpoint,
+            string data
+        )
+        {
+            var request = BuildRequestMessage(apiKey, endpoint, data);
+
+            var response = await _httpClient.SendAsync(request);
```
msohailhussain, line 132: same here, direct return.

OptimizelySDK/Odp/RestApiManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        /// <summary>
+        /// Build an HTTP request message
+        /// </summary>
+        /// <param name="apiKey">ODP public key</param>
+        /// <param name="endpoint">Fully-qualified ODP REST API endpoint</param>
+        /// <param name="data">JSON string version of ODP event data</param>
+        /// <returns>Ready request message to send</returns>
+        private static HttpRequestMessage BuildRequestMessage(string apiKey, string endpoint,
```
msohailhussain, line 144: What's the point to add this method. GraphQL is almost same. Not sure, please explain, couldn't catch if any diff in this method.

OptimizelySDK.Tests/OdpTests/HttpClientTestUtil.cs
```diff
@@ -0,0 +1,70 @@
...
+using Moq.Protected;
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace OptimizelySDK.Tests.OdpTests
+{
+    public static class HttpClientTestUtil
```
msohailhussain, line 27: documentation?

msohailhussain (changes requested):
not reviewed unit test. partial review to unblock if need to work on the feedback.

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -30,62 +34,99 @@ namespace OptimizelySDK.Odp
...
 
-                _logger.Log(LogLevel.WARN, $"Audience segments fetch failed ({errors})");
+                _logger.Log(LogLevel.ERROR, $"{AUDIENCE_FETCH_FAILURE_MESSAGE} ({errors})");
 
-                return emptySegments;
+                return null;
             }
 
-            if (parsedSegments?.Data?.Customer?.Audiences?.Edges is null)
+            if (parsedSegments.Data?.Customer?.Audiences?.Edges is null)
```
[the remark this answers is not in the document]
mikechu-optimizely: I think `is null` performs similarly to javascript's `=== null` ensuring type equality. _I think_. 

With the null-conditional operator chaining if something's `null` I'm validating `null === null` 

I don't think we do any `==` overloading.  I can change it to `== null` as you suggest.

OptimizelySDK/Odp/RestApiManager.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OptimizelySDK.Odp
+{
+    /// <summary>
+    /// Manager for communicating with ODP's REST API endpoint
+    /// </summary>
+    public class RestApiManager : IRestApiManager
```
jaeopt, line 32: Can we change the name: RestApiManager -> OdpEventApiManager

jaeopt (changes requested):
Looks good. A few changes suggested.

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -94,15 +135,101 @@ public GraphQLManager(IErrorHandler errorHandler = null, ILogger logger = null,
...
 
         /// <summary>
-        /// Parses JSON response
+        /// Converts the current QuerySegmentsParameters into a GraphQL query string
         /// </summary>
-        /// <param name="jsonResponse">JSON response from ODP</param>
-        /// <returns>Strongly-typed ODP Response object</returns>
-        public static Response ParseSegmentsResponseJson(string jsonResponse)
+        /// <returns>GraphQL payload</returns>
+        private static string ToGraphQLJson(string userKey, string userValue,
```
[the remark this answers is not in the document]
mikechu-optimizely: Let's talk this over.  For now, I renamed this method `BuildGetSegmentsGraphQLQuery` to show specificity.  

I think using a library (as you suggested) is a good idea.  We'll need to look at finding one that matches this solution's version.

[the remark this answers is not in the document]
mikechu-optimizely: ...or upgrading the solution 😉

OptimizelySDK/Odp/RestApiManager.cs
mikechu-optimizely: Looking at the [Swift reference,](https://github.com/optimizely/swift-sdk/pull/455/files#diff-e050cd60c809ae12d70b80d9474c73b63d5891d6afa08b66b7e900ec13079f4bR89) for the REST API Event Manager, it seems like we're not going to attempt a retry only for server-side errors >= 500.

Perhaps discuss?

OptimizelySDK/Odp/RestApiManager.cs
mikechu-optimizely: There's a bit of a difference in the payload for sending an ODP event to the REST APi and a query to the GraphQL endpoint. 

Happy to walk through the diffs if ya want.  

It was a close similarity.  I'm renaming the two methods for clarity.

OptimizelySDK/Odp/GraphQLManager.cs
```diff
@@ -30,62 +34,99 @@ namespace OptimizelySDK.Odp
...
+            if (segmentsToCheck.Count == 0)
+            {
+                return new string[0];
+            }
 
-            var parameters = new QuerySegmentsParameters.Builder(_logger).WithApiKey(apiKey).
-                WithApiHost(apiHost).WithUserKey(userKey).WithUserValue(userValue).
-                WithSegmentsToCheck(segmentsToCheck).Build();
+            var endpoint = $"{apiHost}/v3/graphql";
+            var query = ToGraphQLJson(userKey.ToString(), userValue, segmentsToCheck);
```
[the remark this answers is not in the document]
mikechu-optimizely: Yes. It is case-sensitive.  Thanks for the catch.

OptimizelySDK/Odp/GraphQLManager.cs
mikechu-optimizely: I think I need some help checking for blocking now that I've changed so much from this review.

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -127,21 +131,21 @@ public bool SendEvents(string apiKey, string apiHost, List<OdpEvent> events)
             string data
         )
         {
-            var request = BuildRequestMessage(apiKey, endpoint, data);
+            var request = BuildOdpEventMessage(apiKey, endpoint, data);
 
-            var response = await _httpClient.SendAsync(request);
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```
jaeopt, line 136: Why do we need override here? Each client may set their own TLS version like 1.3.

jaeopt (changes requested):
Looks good. A clarification regarding TLS.

OptimizelySDK/Odp/OdpEventApiManager.cs
jaeopt: I see. Wondering if we have the same issue with other servers too (datafile, event, and odpEvent) and if we can set "minimum" version to 1.2 instead of fixing the version.

jaeopt (approved):
LGTM

mikechu-optimizely:
> LGTM

Thanks @jaeopt .  Holding for @msohailhussain 's approval 😉

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            var responseStatusCode = response == null ? 0 : (int)response.StatusCode;
+            if (responseStatusCode >= 400)
+            {
+                _logger.Log(LogLevel.ERROR,
+                    $"{EVENT_SENDING_FAILURE_MESSAGE} (${responseStatusCode})");
+            }
+
+            if (responseStatusCode >= 500)
```
msohailhussain, line 115: just curious when will come in this block?

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            catch (Exception ex)
+            {
+                _errorHandler.HandleError(ex);
+                _logger.Log(LogLevel.ERROR, $"{EVENT_SENDING_FAILURE_MESSAGE} (network error)");
+                shouldRetry = true;
+            }
+
+            var responseStatusCode = response == null ? 0 : (int)response.StatusCode;
+            if (responseStatusCode >= 400)
```
msohailhussain, line 109: add a comment why we are not setting retry?

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <param name="endpoint">Fully-qualified ODP REST API endpoint</param>
+        /// <param name="data">JSON string version of ODP event data</param>
+        /// <returns>HTTP response endpoint</returns>
+        private async Task<HttpResponseMessage> SendEventsAsync(string apiKey, string endpoint,
+            string data
+        )
+        {
+            var request = BuildOdpEventMessage(apiKey, endpoint, data);
+
+            return await _httpClient.SendAsync(request);
```
msohailhussain, line 136: we don't need `await` and `async`, directly return _httpClient.SendAsync

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                RequestUri = new Uri(endpoint),
+                Method = HttpMethod.Post,
+                Headers =
+                {
+                    {
+                        Constants.HEADER_API_KEY, apiKey
+                    },
+                },
+                Content = new StringContent(data, Encoding.UTF8, "application/json"),
```
msohailhussain, line 160: for "application/json" can you please add constant.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+        /// <param name="logger">Collect and record events to log</param>
+        /// <param name="errorHandler">Handler to record exceptions</param>
+        /// <param name="httpClient">HttpClient to use to send queries to ODP</param>
+        public OdpSegmentApiManager(ILogger logger = null, IErrorHandler errorHandler = null,
+            HttpClient httpClient = null
+        )
+        {
+            _logger = logger ?? new NoOpLogger();
+            _errorHandler = errorHandler ?? new NoOpErrorHandler();
+            _httpClient = httpClient ?? new HttpClient();
```
msohailhussain, line 70: My recommendation is to create an static method that will return `HttpClient`, should be use across all our SDKs. May be not now but add a ticket in jira to have a single source to return HttpClient in case HttpClient is not provided. 

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.7#examples
```
HttpClient is intended to be instantiated once per application, rather than per-use.
```

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+                    $"{AUDIENCE_FETCH_FAILURE_MESSAGE} (Parameters apiKey or apiHost invalid)");
+                return null;
+            }
+
+            if (segmentsToCheck.Count == 0)
+            {
+                return new string[0];
+            }
+
+            var endpoint = $"{apiHost}/v3/graphql";
```
msohailhussain, line 98: can you please add it as const.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            var query =
+                BuildGetSegmentsGraphQLQuery(userKey.ToString().ToLower(), userValue,
+                    segmentsToCheck);
+
+            var segmentsResponseJson = QuerySegments(apiKey, endpoint, query);
+            if (CanBeJsonParsed(segmentsResponseJson))
```
msohailhussain, line 104: What about renaming it to 
`ValidateJson`

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                _logger.Log(LogLevel.ERROR,
+                    $"{AUDIENCE_FETCH_FAILURE_MESSAGE} (network error)");
+                return null;
+            }
+
+            var parsedSegments = ParseSegmentsResponseJson(segmentsResponseJson);
```
msohailhussain, line 111: I think combine CanBeJsonParsed and ParseSegmentsResponseJson.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            if (parsedSegments is null)
```
msohailhussain, line 112: parsedSegments == `nul` 

`is` is not the right-fit.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            if (parsedSegments.Data?.Customer?.Audiences?.Edges == null)
+            {
+                _logger.Log(LogLevel.ERROR, $"{AUDIENCE_FETCH_FAILURE_MESSAGE} (decode error)");
+
+                return null;
+            }
+
+            return parsedSegments.Data.Customer.Audiences.Edges.
```
msohailhussain, line 135: I think the usual practice is to always use `.` as a prefix in the start of a new line. not as a suffix in the line.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+        /// </summary>
+        /// <param name="userKey">'vuid' or 'fs_user_id key'</param>
+        /// <param name="userValue">Associated value to query for the user key</param>
+        /// <param name="segmentsToCheck">Audience segments to check for experiment inclusion</param>
+        /// <returns>GraphQL string payload</returns>
+        private static string BuildGetSegmentsGraphQLQuery(string userKey, string userValue,
+            IEnumerable segmentsToCheck
+        )
+        {
+            var query =
```
msohailhussain, line 151: Please check this out.
```
            string query =
                @$"
query($userId: String, $audiences: [String]) {{
    {{
        customer({userKey}: $userId) {{
            audiences(subset: $audiences) {{
                edges {{
                    node {{
                        name
                        state
                    }}
                }}
            }}
        }}
    }}
}}";
```

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+                "query($userId: String, $audiences: [String]) {customer(|userKey|: $userId){audiences(subset: $audiences) {edges {node {name state}}}}}".
+                    Replace("|userKey|", userKey);
+
+            var variables =
```
msohailhussain, line 155: use same as above.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                RequestUri = new Uri(endpoint),
+                Method = HttpMethod.Post,
+                Headers =
+                {
+                    {
+                        Constants.HEADER_API_KEY, apiKey
+                    },
+                },
+                Content = new StringContent(query, Encoding.UTF8, "application/json"),
```
msohailhussain, line 234: `application/json` should be `const`

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+
+            return request;
+        }
+
+        /// <summary>
+        /// Ensure a string has content that can be parsed from JSON to an object
+        /// </summary>
+        /// <param name="jsonToValidate">Value containing possible JSON</param>
+        /// <returns>True if content could be interpreted as JSON else False</returns>
+        private static bool CanBeJsonParsed(string jsonToValidate)
```
msohailhussain, line 245: name doesn't align with the functionality. Don't need this method, can be used as a condition in `ParseSegmentsResponseJson`

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -0,0 +1,260 @@
...
+        {
+            return string.IsNullOrWhiteSpace(jsonToValidate);
+        }
+
+        /// <summary>
+        /// Parses JSON response
+        /// </summary>
+        /// <param name="jsonResponse">JSON response from ODP</param>
+        /// <returns>Strongly-typed ODP Response object</returns>
+        public static Response ParseSegmentsResponseJson(string jsonResponse)
```
msohailhussain, line 255: `UnMarshal` or `Deserialize` may be suitable here instead of `Parse`

msohailhussain (changes requested):
please address few more questions.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
mikechu-optimizely: New Jira issue created to address in a later effort.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -132,62 +131,85 @@ public class OdpSegmentApiManager : IOdpSegmentApiManager
...
+                        }
+                    }"", 
+                    ""variables"" : {
+                        ""userId"": ""{userValue}"",
+                        ""audiences"": {audiences}
+                    }
+                }"
+                    .Replace("{userKey}", userKey)
+                    .Replace("{userValue}", userValue)
+                    .Replace("{audiences}", JsonConvert.SerializeObject(segmentsToCheck));
```
msohailhussain, line 176: please use strings.Join instead of serialization.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -95,21 +94,21 @@ public class OdpSegmentApiManager : IOdpSegmentApiManager
...
+            if (string.IsNullOrWhiteSpace(segmentsResponseJson))
             {
                 _logger.Log(LogLevel.ERROR,
                     $"{AUDIENCE_FETCH_FAILURE_MESSAGE} (network error)");
                 return null;
             }
 
-            var parsedSegments = ParseSegmentsResponseJson(segmentsResponseJson);
-            if (parsedSegments is null)
+            var parsedSegments = DeserializeSegmentsFromJson(segmentsResponseJson);
```
msohailhussain, line 110: change hte variable name.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -231,28 +253,19 @@ string query
...
-            return string.IsNullOrWhiteSpace(jsonToValidate);
-        }
-
         /// <summary>
         /// Parses JSON response
         /// </summary>
         /// <param name="jsonResponse">JSON response from ODP</param>
         /// <returns>Strongly-typed ODP Response object</returns>
-        public static Response ParseSegmentsResponseJson(string jsonResponse)
+        public static Response DeserializeSegmentsFromJson(string jsonResponse)
```
msohailhussain, line 268: Can you please add a test case if not written, in providing invalid json, what will be the behavior?

OptimizelySDK/Odp/OdpSegmentApiManager.cs
```diff
@@ -132,62 +131,85 @@ public class OdpSegmentApiManager : IOdpSegmentApiManager
...
-                "query($userId: String, $audiences: [String]) {customer(|userKey|: $userId){audiences(subset: $audiences) {edges {node {name state}}}}}".
-                    Replace("|userKey|", userKey);
-
-            var variables =
-                "\"variables\": { \"userId\": \"|userValue|\", \"audiences\": |audiences| }".
-                    Replace("|userValue|", userValue).
-                    Replace("|audiences|", JsonConvert.SerializeObject(segmentsToCheck));
-
-            return $"{{ \"query\": \"{query}\", {variables} }}";
+            return
```
msohailhussain, line 151: I will highly recommend to sanitize all arguments, since we are appending string. It may inject any runnable code.

msohailhussain:
@mikechu-optimizely Overall your PR looks good, please address 4 more questions and then lgtm.

msohailhussain:
last but not least please fix Lint issues.

OptimizelySDK/Odp/OdpSegmentApiManager.cs
mikechu-optimizely: I think the Join() to create a `["string_1", "string_2"]` will not read as well.

What do you think?

mikechu-optimizely:
> last but not least please fix Lint issues.

It seems like the lint error here in the [GitHub Action](https://github.com/optimizely/csharp-sdk/actions/runs/3322188387/jobs/5490861536#step:4:46) is about a missing auth token.  

Would you like me to pause the Pull Request and resolve this step so we can make the Csharp CI with .NET / Lint Code Base (pull_request) Failing after 2m required?

msohailhussain:
this is fine for me, we should resolve token issue in future.

msohailhussain (approved):
lgtm.

PR #319, opened 2022-11-14
chore: Fix and Reconfigure GitHub Actions

mikechu-optimizely:
## Summary
- Fix failing lint_code_base
- Change to use Windows image to compile SDK

## Test plan
- Verify all GitHub Actions are passing

## Issues
- FSSDK-8681

### DIFF
--- a/.github/workflows/csharp.yml
+++ b/.github/workflows/csharp.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Restore NuGet packages
         run: nuget restore ./OptimizelySDK.Travis.sln
       - name: Build solution
-        run: msbuild /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=$(pwd)/keypair.snk /p:Configuration=Release ./OptimizelySDK.sln
+        run: msbuild /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=$(pwd)/keypair.snk /p:Configuration=Release ./OptimizelySDK.Travis.sln
       - name: Install NUnit Console
         run: nuget install NUnit.Console -Version 3.15.2 -DirectDownload -OutputDirectory .
       - name: Run NUnit tests
--- a/.github/workflows/csharp.yml
+++ b/.github/workflows/csharp.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   RELEASE_BRANCH: "master"
-  WINDOWS_2019_SN_PATH: "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe"
+  WINDOWS_2019_SN_PATH: C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe
 
 jobs:
   lint_code_base:
--- a/.github/workflows/csharp.yml
+++ b/.github/workflows/csharp.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   RELEASE_BRANCH: "master"
+  WINDOWS_2019_SN_PATH: "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe"
 
 jobs:
   lint_code_base:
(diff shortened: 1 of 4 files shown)


mikechu-optimizely:
@msohailhussain Please check that I've set the correct "Required" tasks.

.github/workflows/csharp.yml
```diff
@@ -52,35 +56,40 @@ jobs:
...
-          mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll
-      - name: find and sign dll
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1
+      - name: Setup NuGet
+        uses: NuGet/setup-nuget@v1
+      - name: Restore NuGet packages
+        run: nuget restore ./OptimizelySDK.Travis.sln
```
msohailhussain, line 66: Can you please run OptimizelySDK.sln instead of OptimizelySDK.Travis.sln

.github/workflows/csharp.yml
```diff
@@ -52,35 +56,40 @@ jobs:
...
+      - name: Build solution
+        run: msbuild /p:SignAssembly=true /p:AssemblyOriginatorKeyFile=$(pwd)/keypair.snk /p:Configuration=Release ./OptimizelySDK.Travis.sln
+      - name: Install NUnit Console
```
msohailhussain, line 69: Please move it before running msbuild. SO all setups can be seen together.

.github/workflows/csharp.yml
```diff
@@ -52,35 +56,40 @@ jobs:
...
+        run: nuget install NUnit.Console -Version 3.15.2 -DirectDownload -OutputDirectory .
+      - name: Run NUnit tests
+        run: ./NUnit.ConsoleRunner.3.15.2\tools\nunit3-console.exe /timeout 10000 /process Separate ./OptimizelySDK.Tests/bin/Release/OptimizelySDK.Tests.dll
```
msohailhussain, line 72: Good to have, if you can specify each parameter what it does. 
e.g. /process Separate and /timeout, or add a link to see what parameters are supported.

.github/workflows/csharp.yml
```diff
@@ -52,35 +56,40 @@ jobs:
...
+      - name: Find and sign all DLLs
         id: unit_tests
         run: |
-          sudo find . -path './OptimizelySDK*bin/Release/OptimizelySDK*.dll' -not -regex '.*Tests.*' -print0 | while IFS= read -r -d '' file; do sn -R $file ./keypair.snk; done
-      - name: Check on success
+          $sn = "C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe";
```
msohailhussain, line 76: great job. can you please find a way if there is any variable to represent base path. don't hardcode NETFX4.8 path.

msohailhussain (approved):
all looks good. please address my comments and then merge the PR.

.github/workflows/csharp.yml
mikechu-optimizely: Sure. That makes sense.  

I intended to only execute installations before they were needed in case of failure.  eg If build failed, no need to install NUnit.

.github/workflows/csharp.yml
mikechu-optimizely: Oh, that's a great idea. Yes. I'll add a comment.

.github/workflows/csharp.yml
mikechu-optimizely: Thanks.

I struggled to enumerate the path to copy of `sn.exe`.  There are several copies of it on the `windows-2019` image.  The PowerShell to find and select only one is not clear to the reader, so I decided to hard-code the path.  

I believe leaving this hard-coded poses little risk. The Microsoft's image *should* (in theory) not change since it's shared among so many builds.

I do think it's a great idea to move this base path to a YAML level variable, so it's prominent and if needed can be updated.

.github/workflows/csharp.yml
mikechu-optimizely: 🤔 Thinking more on this… I'd like to think green and code sustainably here.  

By performing the installation only _if_ and just before it's needed, we're reducing the amount of resources (albeit small) being used on a potentially unnecessary step.  

If you feel strongly over this re-ordering, please DM me on Teams, and I'll roll it into another PR.

.github/workflows/csharp.yml
mikechu-optimizely: The `OptimizelySDK.Travis.sln` does not include the demo app OptimizelySDK.DemoApp.csproj which has [build errors](https://github.com/optimizely/csharp-sdk/actions/runs/3470770177/jobs/5799387724#step:6:53).  

I'm guessing the demo app was not needed, so someone created the Travis version.  

In later sprints, I think we should fix or remove the demo app.

It's likely safe to continue to build the Travis solution instead.

PR #322, opened 2022-11-29
feat: Added ODPManager implementation

mikechu-optimizely:
## Summary
Added ODPManager Implementation which does the following.
1. Initializes and provides access to ODPEventManager and ODPSegmentManager
2. Provides updated ODPConfig settings to event manager and segment manager.
3. Stops Event Manager thread when closed.

## Test plan
1. Manually tested thoroughly
2. Added unit tests.

## Issues
FSSDK-8453

### DIFF
--- a/OptimizelySDK/Odp/IOdpManager.cs
+++ b/OptimizelySDK/Odp/IOdpManager.cs
@@ -60,6 +60,6 @@ namespace OptimizelySDK.Odp
         /// <summary>
         /// Sends signal to stop Event Manager and clean up ODP Manager use
         /// </summary>
-        void Close();
+        void Dispose();
     }
 }
--- a/OptimizelySDK/Odp/IOdpManager.cs
+++ b/OptimizelySDK/Odp/IOdpManager.cs
@@ -38,7 +38,7 @@ namespace OptimizelySDK.Odp
         /// <param name="userId">FS User ID</param>
         /// <param name="options">Options used during segment cache handling</param>
         /// <returns>Qualified segments for the user from the cache or the ODP server</returns>
-        List<string> FetchQualifiedSegments(string userId, List<OdpSegmentOption> options);
+        string[] FetchQualifiedSegments(string userId, List<OdpSegmentOption> options);
 
         /// <summary>
         /// Send identification event to ODP for a given full-stack User ID
--- a/OptimizelySDK/Odp/IOdpManager.cs
+++ b/OptimizelySDK/Odp/IOdpManager.cs
@@ -18,18 +18,48 @@ using System.Collections.Generic;
 
 namespace OptimizelySDK.Odp
 {
+    /// <summary>
+    /// Interface describing orchestration of segment manager, event manager, and ODP config
+    /// </summary>
     public interface IOdpManager
     {
+        /// <summary>
+        /// Update the settings being used for ODP configuration and reset/restart dependent processes
+        /// </summary>
+        /// <param name="apiKey">Public API key from ODP</param>
+        /// <param name="apiHost">Host portion of the URL of ODP</param>
+        /// <param name="segmentsToCheck">Audience segments to consider</param>
+        /// <returns>True if settings were update otherwise False</returns>
         bool UpdateSettings(string apiKey, string apiHost, List<string> segmentsToCheck);
 
+        /// <summary>
+        /// Attempts to fetch and return a list of a user's qualified segments.
+        /// </summary>
+        /// <param name="userId">FS User ID</param>
+        /// <param name="options">Options used during segment cache handling</param>
+        /// <returns>Qualified segments for the user from the cache or the ODP server</returns>
         List<string> FetchQualifiedSegments(string userId, List<OdpSegmentOption> options);
 
+        /// <summary>
+        /// Send identification event to ODP for a given full-stack User ID
+        /// </summary>
+        /// <param name="userId">User ID to send</param>
         void IdentifyUser(string userId);
 
+        /// <summary>
+        /// Add event to queue for sending to ODP
+        /// </summary>
+        /// <param name="type">Type of event (typically `fullstack` from server-side SDK events)</param>
+        /// <param name="action">Subcategory of the event type</param>
+        /// <param name="identifiers">Key-value map of user identifiers</param>
+        /// <param name="data">Event data in a key-value pair format</param>
         void SendEvent(string type, string action, Dictionary<string, string> identifiers,
             Dictionary<string, object> data
         );
 
+        /// <summary>
+        /// Sends signal to stop Event Manager and clean up ODP Manager use
+        /// </summary>
         void Close();
     }
 }
--- a/OptimizelySDK/Odp/IOdpManager.cs
+++ b/OptimizelySDK/Odp/IOdpManager.cs
@@ -1,0 +1,35 @@
+﻿/* 
+ * Copyright 2022 Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Collections.Generic;
+
+namespace OptimizelySDK.Odp
+{
+    public interface IOdpManager
+    {
+        bool UpdateSettings(string apiKey, string apiHost, List<string> segmentsToCheck);
+
+        List<string> FetchQualifiedSegments(string userId, List<OdpSegmentOption> options);
+
+        void IdentifyUser(string userId);
+
+        void SendEvent(string type, string action, Dictionary<string, string> identifiers,
+            Dictionary<string, object> data
+        );
+
+        void Close();
+    }
+}
--- a/OptimizelySDK/Odp/OdpConfig.cs
+++ b/OptimizelySDK/Odp/OdpConfig.cs
@@ -25,59 +25,25 @@ namespace OptimizelySDK.Odp
         /// <summary>
         /// Public API key for the ODP account from which the audience segments will be fetched (optional).
         /// </summary>
-        private volatile string _apiKey;
-
-        public string ApiKey
-        {
-            get
-            {
-                return _apiKey;
-            }
-            private set
-            {
-                _apiKey = value;
-            }
-        }
+        public string ApiKey { get; private set; }
 
         /// <summary>
         /// Host of ODP audience segments API.
         /// </summary>
-        private volatile string _apiHost;
-
-        public string ApiHost
-        {
-            get
-            {
-                return _apiHost;
-            }
-            private set
-            {
-                _apiHost = value;
-            }
-        }
+        public string ApiHost { get; private set; }
 
         /// <summary>
         /// All ODP segments used in the current datafile (associated with apiHost/apiKey).
         /// </summary>
-        private volatile List<string> _segmentsToCheck;
+        public List<string> SegmentsToCheck { get; private set; }
 
-        public List<string> SegmentsToCheck
+
+        public OdpConfig(string apiKey = null, string apiHost = null,
+            List<string> segmentsToCheck = null
+        )
         {
-            get
-            {
-                return _segmentsToCheck;
-            }
-
-            private set
-            {
-                _segmentsToCheck = value;
-            }
-        }
-
-        public OdpConfig(string apiKey, string apiHost, List<string> segmentsToCheck)
-        {
-            ApiKey = apiKey;
-            ApiHost = apiHost;
+            ApiKey = apiKey ?? string.Empty;
+            ApiHost = apiHost ?? string.Empty;
             SegmentsToCheck = segmentsToCheck ?? new List<string>(0);
         }
 
--- a/OptimizelySDK/Odp/OdpConfig.cs
+++ b/OptimizelySDK/Odp/OdpConfig.cs
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+using System;
 using System.Collections.Generic;
 
 namespace OptimizelySDK.Odp
@@ -116,6 +117,14 @@ namespace OptimizelySDK.Odp
         public bool HasSegments()
         {
             return SegmentsToCheck?.Count > 0;
+        }
+
+        public bool Equals(OdpConfig toCompare)
+        {
+            return ApiKey.Equals(toCompare.ApiKey, StringComparison.OrdinalIgnoreCase) &&
+                   ApiHost.Equals(toCompare.ApiHost, StringComparison.OrdinalIgnoreCase) &&
+                   SegmentsToCheck.TrueForAll(
+                       segment => toCompare.SegmentsToCheck.Contains(segment));
         }
     }
 }
--- a/OptimizelySDK/Odp/OdpConfig.cs
+++ b/OptimizelySDK/Odp/OdpConfig.cs
@@ -82,27 +82,6 @@ namespace OptimizelySDK.Odp
         }
 
         /// <summary>
-        /// Update the ODP configuration details
-        /// </summary>
-        /// <param name="apiKey">Public API key for the ODP account</param>
-        /// <param name="apiHost">Host of ODP audience segments API</param>
-        /// <param name="segmentsToCheck">Audience segments</param>
-        /// <returns>true if configuration was updated successfully otherwise false</returns>
-        public virtual bool Update(string apiKey, string apiHost, List<string> segmentsToCheck)
-        {
-            if (ApiKey == apiKey && ApiHost == apiHost && SegmentsToCheck == segmentsToCheck)
-            {
-                return false;
-            }
-
-            ApiKey = apiKey;
-            ApiHost = apiHost;
-            SegmentsToCheck = segmentsToCheck;
-
-            return true;
-        }
-
-        /// <summary>
         /// Determines if ODP configuration has the minimum amount of information
         /// </summary>
         /// <returns>true if ODP configuration can be used otherwise false</returns>
--- a/OptimizelySDK/Odp/OdpConfig.cs
+++ b/OptimizelySDK/Odp/OdpConfig.cs
@@ -16,10 +16,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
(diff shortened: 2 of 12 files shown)


OptimizelySDK/OptimizelySDK.csproj
```diff
@@ -101,8 +101,9 @@
     <Compile Include="Odp\Entity\OdpEvent.cs" />
     <Compile Include="Odp\Entity\Response.cs" />
     <Compile Include="Odp\Enums.cs" />
+    <Compile Include="Odp\IOdpManager.cs" />
     <Compile Include="Odp\IOdpSegmentManager.cs" />
-    <Compile Include="Odp\OdpConfig.cs" />
```
mikechu-optimizely, line 105: This was already declared on line 112/113 below.  Added  `OdpManager.cs` in its place.

OptimizelySDK/Odp/IOdpManager.cs
```diff
@@ -0,0 +1,65 @@
...
+    public interface IOdpManager
+    {
+        /// <summary>
+        /// Update the settings being used for ODP configuration and reset/restart dependent processes
+        /// </summary>
+        /// <param name="apiKey">Public API key from ODP</param>
+        /// <param name="apiHost">Host portion of the URL of ODP</param>
+        /// <param name="segmentsToCheck">Audience segments to consider</param>
+        /// <returns>True if settings were update otherwise False</returns>
+        bool UpdateSettings(string apiKey, string apiHost, List<string> segmentsToCheck);
```
msohailhussain, line 33: shouldn't it be OdpConfig in args

OptimizelySDK/Odp/IOdpManager.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Attempts to fetch and return a list of a user's qualified segments.
+        /// </summary>
+        /// <param name="userId">FS User ID</param>
+        /// <param name="options">Options used during segment cache handling</param>
+        /// <returns>Qualified segments for the user from the cache or the ODP server</returns>
+        List<string> FetchQualifiedSegments(string userId, List<OdpSegmentOption> options);
```
msohailhussain, line 41: nit: This should be `Array`

OptimizelySDK/Odp/IOdpManager.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="identifiers">Key-value map of user identifiers</param>
+        /// <param name="data">Event data in a key-value pair format</param>
+        void SendEvent(string type, string action, Dictionary<string, string> identifiers,
+            Dictionary<string, object> data
+        );
+
+        /// <summary>
+        /// Sends signal to stop Event Manager and clean up ODP Manager use
+        /// </summary>
+        void Close();
```
msohailhussain, line 63: This should be Dispose. My understanding is if it's closed, then can't use again.

OptimizelySDK/Odp/OdpConfig.cs
```diff
@@ -117,5 +118,18 @@ public bool HasSegments()
         {
             return SegmentsToCheck?.Count > 0;
         }
+
+        /// <summary>
+        /// Determine equality between two OdpConfig objects
+        /// </summary>
+        /// <param name="toCompare">OdpConfig object to compare current instance against</param>
+        /// <returns>True if equal otherwise False</returns>
+        public bool Equals(OdpConfig toCompare)
```
msohailhussain, line 127: what's the reason to add this method. This will raise exception if ApiKey is null. Secondly use equals method where needed in this class and implement IEqu... interface as well.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// Update the settings being used for ODP configuration and reset/restart dependent processes
+        /// </summary>
+        /// <param name="apiKey">Public API key from ODP</param>
+        /// <param name="apiHost">Host portion of the URL of ODP</param>
+        /// <param name="segmentsToCheck">Audience segments to consider</param>
+        /// <returns>True if settings were update otherwise False</returns>
+        public bool UpdateSettings(string apiKey, string apiHost, List<string> segmentsToCheck)
+        {
+            var newConfig = new OdpConfig(apiKey, apiHost, segmentsToCheck);
+            if (_odpConfig.Equals(newConfig))
```
msohailhussain, line 65: can't we use `==`

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,274 @@
...
+            {
+                return false;
+            }
+
+            _odpConfig = newConfig;
+
+            EventManager.Flush();
```
msohailhussain, line 72: add comments in a code, why `flushing`

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        /// <summary>
+        /// Attempts to fetch and return a list of a user's qualified segments.
+        /// </summary>
+        /// <param name="userId">FS User ID</param>
+        /// <param name="options">Options used during segment cache handling</param>
+        /// <returns>Qualified segments for the user from the cache or the ODP server</returns>
+        public List<string> FetchQualifiedSegments(string userId, List<OdpSegmentOption> options)
+        {
+            if (!_enabled || SegmentManager == null || !_odpConfig.IsReady())
```
msohailhussain, line 89: a minor comment, 
` SegmentManager == null || !_enabled || !_odpConfig.IsReady()`

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// </summary>
+        /// <param name="type">Type of event (typically `fullstack` from server-side SDK events)</param>
+        /// <param name="action">Subcategory of the event type</param>
+        /// <param name="identifiers">Key-value map of user identifiers</param>
+        /// <param name="data">Event data in a key-value pair format</param>
+        public void SendEvent(string type, string action, Dictionary<string, string> identifiers,
+            Dictionary<string, object> data
+        )
+        {
+            if (!_enabled || EventManager == null || !_odpConfig.IsReady())
```
msohailhussain, line 131: This validations should be in a single method and return true or false. please refactor it.

msohailhussain (changes requested):
please address.

OptimizelySDK/Odp/OdpConfig.cs
mikechu-optimizely: The goal of this is to control the equality determination of two ODP Configs. 

Thanks for the callout on null exception.  I'll change to use `IEqualityComparer`.

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: Yes Sir.  I'll update once using the `IEqualityComparer` from `OdpConfig`

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: I can re-arrange (I think that's what you're asking)  

I thought that perhaps `!_enabled` would evaluate to true more often than the others, but happy to switch.

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: My copy-paste finger started hurting, and I thought of that too.  

L89, would have me injecting the Manager that should be checked for `null` so I just left it.

Let me play with some options....

OptimizelySDK/Odp/IOdpManager.cs
mikechu-optimizely: Oh, yes. Sorry.  Old habit from previous work.

mikechu-optimizely (commented):
Thanks for the comments @msohailhussain . 

I'll hold for @jaeopt 's too before pushing to the repo

OptimizelySDK/Odp/OdpConfig.cs
mikechu-optimizely: Oh maybe `IEquatable<T>` might work instead of having a comparer. 🤔

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: // because @jaeopt did it in [Swift](https://github.com/optimizely/swift-sdk/blob/master/Sources/ODP/OdpManager.swift#L128) 😁

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+                return false;
+            }
+
+            _odpConfig = newConfig;
+
+            // flush old events using old odp publicKey (if exists) before updating odp key.
+            // NOTE: It should be rare but possible that odp public key is changed for the same datafile (sdkKey).
+            // - Try to send all old events with the previous public key.
+            // - If it fails to flush all the old events here (network error), remaining events will be discarded.
+            EventManager.Flush();
```
jaeopt, line 76: How can we guarantee all pending events use the old ODPConfig with the non-blocking flush?

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+        /// <returns>True if settings were update otherwise False</returns>
+        public bool UpdateSettings(string apiKey, string apiHost, List<string> segmentsToCheck)
+        {
+            var newConfig = new OdpConfig(apiKey, apiHost, segmentsToCheck);
+            if (_odpConfig.Equals(newConfig))
+            {
+                return false;
+            }
+
+            _odpConfig = newConfig;
```
jaeopt, line 70: When ODPConfig is updated, the OdpConfig object itself is replaced in ODPManager, while copying fields in ODPEventManager and ODPSegmentManager. Can we make it consistent in either way?

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+                _logger.Log(LogLevel.DEBUG, "ODP identify event not dispatched (ODP disabled).");
+                return;
+            }
+
+            if (!EventManager.IsStarted)
+            {
+                _logger.Log(LogLevel.DEBUG,
+                    "ODP identify event not dispatched (ODP not integrated).");
+                return;
+            }
```
jaeopt, line 119: We can drop this. It should be rare and it'll be filtered out in EventManager anyway.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+        /// <param name="action">Subcategory of the event type</param>
+        /// <param name="identifiers">Key-value map of user identifiers</param>
+        /// <param name="data">Event data in a key-value pair format</param>
+        public void SendEvent(string type, string action, Dictionary<string, string> identifiers,
+            Dictionary<string, object> data
+        )
+        {
+            if (EventManagerOrConfigNotReady())
+            {
+                _logger.Log(LogLevel.DEBUG, "ODP event not dispatched (ODP disabled).");
```
jaeopt, line 137:
```suggestion
                _logger.Log(LogLevel.ERROR, "ODP event not dispatched (ODP disabled).");
```
Implicit event sending (IdentifyUser) should be DEBUG level, but when clients try to send events when ODP is not ready, it should be ERROR.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+                return;
+            }
+
+            if (!EventManager.IsStarted)
+            {
+                _logger.Log(LogLevel.DEBUG,
+                    "ODP event not dispatched (ODP not integrated).");
+                return;
+            }
```
jaeopt, line 146: We can drop this too.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+            /// <summary>
+            /// Build OdpManager instance using collected parameters
+            /// </summary>
+            /// <param name="asEnabled">Should mark as enabled upon initialization</param>
+            /// <returns>OdpManager instance</returns>
+            public OdpManager Build(bool asEnabled = true)
+            {
+                var manager = new OdpManager
+                {
+                    _odpConfig = _odpConfig,
```
jaeopt, line 235: _odpConfig can be null if not set `withOdpConfig`. Is it safe?

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -0,0 +1,296 @@
...
+                    _logger = _logger ?? new NoOpLogger(),
+                    _enabled = asEnabled,
+                };
+
+                _errorHandler = _errorHandler ?? new NoOpErrorHandler();
+
+                if (_eventManager == null)
```
jaeopt, line 242: if _enabled == false, then we do not need instantiate EventManager and SegmentManager.

jaeopt (changes requested):
Looks good. A few changes suggested.

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: That's a good consideration. I think this `.Flush()` should grab a copy of the `_odpConfig` in the `OdpEventManager` and inject it into the `_flushSignal` object.  The secondary thread would then have the config to use for flush requests.

Does this seem like a suitable solution, or is this a corner case?  
How does Swift (or Python) handle this situation?
What's your direction?

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: I put a check in the `Build()` for you 😄

msohailhussain (approved):
lgtm

OptimizelySDK/Odp/OdpConfig.cs
```diff
@@ -81,27 +81,6 @@ public OdpConfig(string apiKey, string apiHost, List<string> segmentsToCheck)
             SegmentsToCheck = segmentsToCheck ?? new List<string>(0);
         }
 
-        /// <summary>
```
jaeopt, line 82: If the entire odpconfig is replaced for update, then we do not need t"volitile" for internal fields.

OptimizelySDK/Odp/OdpEventManager.cs
```diff
@@ -349,7 +349,7 @@ public void IdentifyUser(string userId)
         /// <param name="odpConfig">Configuration object containing new values</param>
         public void UpdateSettings(OdpConfig odpConfig)
         {
-            _odpConfig.Update(odpConfig.ApiKey, odpConfig.ApiHost, odpConfig.SegmentsToCheck);
+            _odpConfig = odpConfig;
```
jaeopt, line 352: If you use to replace _odpConfig, don't we need to set it to "volitile"?

OptimizelySDK/Odp/OdpSegmentManager.cs
```diff
@@ -149,7 +149,7 @@ private static string GetCacheKey(string userKey, string userValue)
         /// <param name="odpConfig">New ODP Configuration to apply</param>
         public void UpdateSettings(OdpConfig odpConfig)
         {
-            _odpConfig.Update(odpConfig.ApiKey, odpConfig.ApiHost, odpConfig.SegmentsToCheck);
+            _odpConfig = odpConfig;
```
jaeopt, line 152: volitile

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -248,20 +216,33 @@ public Builder WithCacheImplementation(ICache<List<string>> cache)
             /// <returns>OdpManager instance</returns>
             public OdpManager Build(bool asEnabled = true)
             {
+                _logger = _logger ?? new DefaultLogger();
+                _errorHandler = _errorHandler ?? new NoOpErrorHandler();
+
+                if (_odpConfig == null)
```
jaeopt, line 222: What about setting `if (odpConfig == null) _odpConfig = new ODPConfig()` default. 
Initially that's the best they can do when initializing ODPManager.

OptimizelySDK/Odp/OdpManager.cs
jaeopt: @mikechu-optimizely  We can consider -
1. (swift) the current key and host are captures in the flush thread (so they can be used until flushing completed) before updating odpConfig or
2. (python) delay config update until flush is completed with the current key and host.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -248,20 +216,33 @@ public Builder WithCacheImplementation(ICache<List<string>> cache)
...
+                {
+                    return manager;
+                }
 
                 if (_eventManager == null)
                 {
                     var eventApiManager = new OdpEventApiManager(_logger, _errorHandler);
 
-                    manager.EventManager = new OdpEventManager.Builder().WithOdpConfig(_odpConfig).
+                    manager.EventManager = new OdpEventManager.Builder().
```
jaeopt, line 244: Wondering why using builder for OdpEventManager while not for OdpSegmentManager. Looks weird with 2 different patterns for similar objects.

jaeopt (changes requested):
A few more changes suggested.

OptimizelySDK/Odp/OdpManager.cs
mikechu-optimizely: Segment Manager is not too complex to instantiate.  There's not a lot of logic that needed to be in the constructor, so I let a standard constructor do that work. 

With Event Manager and ODP Manager, the combinations of instantiation parameters and logic in those combinations warranted (in my opinion) to use the Builder Pattern.  

In practice, I don't have a hard limit of constructor length before breaking the logic out.

OptimizelySDK/Odp/OdpEventManager.cs
mikechu-optimizely: `volatile` can [only be set on some simple/primitive types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile).  

Replacing the config as a whole now (instead of individual field values on the OdpConfig) should not be good. I'm not seeing where a child thread updates on the config.

OptimizelySDK/Odp/OdpManager.cs
```diff
@@ -35,7 +35,7 @@ public class OdpManager : IOdpManager, IDisposable
         /// <summary>
         /// Configuration used to communicate with ODP
         /// </summary>
-        private volatile OdpConfig _odpConfig;
+        private OdpConfig _odpConfig;
```
jaeopt, line 38: Why do you remove "volitile" here? Multiple thread read/write it here too.

jaeopt (changes requested):
Looks good. One more fix for thread-safe.

jaeopt (approved):
LGTM

PR #323, opened 2022-12-15
feat: Integrated ODPManager with UserContext and Optimizely client

mikechu-optimizely:
## Summary
This PR integrates ODP functionality with UserContext and Optimizely client to make it available for users. There are two ways to integrate ODPManager.

😬 Sorry about the numerous auto-format changes polluting the core enhancements.  Need to satisfy the linter.

## Test plan
1. Manually tested thoroughly
2. Will add unit tests after the first review passes.

## Issues
FSSDK-8476

PR #325, opened 2023-01-23
Fix: Some suggested changes related to ODP

mnoman09:
## Summary
- Changes enum to match with DecideOption naming convention (We are expecting in FSC that it would be same as decideOptions). 
- Flushes odp event whenever updateSetting is called. (I am not sure, if this is required as I couldn't find a place where Updating config will fire previous config remaining events.)
- Adds additional check to avoid null pointer exception.
- Corrects ODP segments query format. As it was throwing an exception of bad request.
- returning null in case of an error. 
- Corrects the condition to add UpdateConfig listener for ODP config. 
- Made a separate function for creating copy of userContext internally, which will not fire sendOdpEvent Initialized by default. 

## Test plan
Tested it with FSC and all the Tests should pass on FSC. 
- Haven't looked at unit tests. 

## Issues
FSSDK-8601


mikechu-optimizely:
Hi @mnoman09 

Would you want to merge this before #323? 
Do you have an FSSDK-#### number for the description?

Thanks.

mikechu-optimizely:
Oh. Ignore the first question, I see you've been merging `mike/odp-usercontext-optimizelyclient`

mikechu-optimizely (approved):
LGTM. I'll be ready to re-approve if needed.

mikechu-optimizely:
Updated initial PR comment > Issues (<h2>) to include FSSDK-8601.  Please update if in correct.

mnoman09:
> Updated initial PR comment > Issues (`<h2>`) to include FSSDK-8601. Please update if in correct.

Yes, these are the suggested changes which are required to pass all the FSC tests. So you can merge it into [mike/odp-usercontext-optimizelyclient](https://github.com/optimizely/csharp-sdk/tree/mike/odp-usercontext-optimizelyclient).

PR #326, opened 2023-02-08
Fix: ODP Event manager not consistent in triggering events.

mnoman09:
## Summary
- Refactored OdpEventManager flushEvent function. 
- To wait for events using flushInterval when eventBatch is not empty.
- Or wait indefinitely if no event is in batch until any event arrive. 
- Avoid calling flush on shutdown if no events are present 
- Avoid calling unnecessary flushQueue if _currentBatch.Count is zero. 
 
## Test plan
Passing build link with FSC: [link](https://github.com/optimizely/fullstack-sdk-compatibility-suite/actions/runs/4163960853/jobs/7206348700)

## Issues
- [TODO]


mikechu-optimizely (approved):
LGTM

PR #337, opened 2023-03-17
[FSSDK-9001] Fix: ODP should be enabled by default when initialized optimizely using OptimizelyFactory  

mnoman09:
## Summary
- Using default odp manager when initializing Optimizely instance using OptimizelyFactory. This will enable the odp by default. 

## Test plan
All tests should pass. 

## Issues
- [FSSDK-9001](https://jira.sso.episerver.net/browse/FSSDK-9001)


mikechu-optimizely (approved):
LGTM.  Thanks!

PR #342, opened 2023-03-24
[FSSDK-8948] fix(odp): check odp identifiers not null or empty before sending

mnoman09:
## Summary
- Check if ODP events have at least one key-value pair for identifiers.
If not, discard the request without sending to the ODP server.

## Test plan
- Add a test with empty identifiers and null identifier.
- All tests should pass. 

## Issues
- [FSSDK-8948](https://jira.sso.episerver.net/browse/FSSDK-8948)

### DIFF
--- a/OptimizelySDK/Odp/Constants.cs
+++ b/OptimizelySDK/Odp/Constants.cs
@@ -1,5 +1,5 @@
 ﻿/* 
- * Copyright 2022 Optimizely
+ * Copyright 2022-2023 Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -65,7 +65,12 @@ namespace OptimizelySDK.Odp
         /// <summary>
         /// Default message to log when an ODP Event contains invalid data
         /// </summary>
-        public const string ODP_INVALID_DATA_MESSAGE = "ODP data is not valid.";
+        public const string ODP_INVALID_DATA_MESSAGE = "ODP event send failed (event data is not valid).";
+
+        /// <summary>
+        /// Default message to log when an ODP Identifiers is invalid
+        /// </summary>
+        public const string ODP_INVALID_IDENTIFIERS_MESSAGE = "ODP event send failed (event identifiers must have at least one key-value pair).";
 
         /// <summary>
         /// Default message to log when sending ODP event fails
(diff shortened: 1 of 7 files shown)


OptimizelySDK/Odp/Constants.cs
```diff
@@ -65,7 +65,12 @@ public static class Constants
         /// <summary>
         /// Default message to log when an ODP Event contains invalid data
         /// </summary>
-        public const string ODP_INVALID_DATA_MESSAGE = "ODP data is not valid.";
+        public const string ODP_INVALID_DATA_MESSAGE = "ODP event send failed (event data is not valid).";
```
mikechu-optimizely, line 68: The TDD has a defined message 
odpInvalidData ERROR "ODP data is not valid."

I think we should adhere unless @jaeopt has other input.

OptimizelySDK/Odp/Constants.cs
```diff
@@ -65,7 +65,12 @@ public static class Constants
...
+
+        /// <summary>
+        /// Default message to log when an ODP Identifiers is invalid
+        /// </summary>
+        public const string ODP_INVALID_IDENTIFIERS_MESSAGE = "ODP event send failed (event identifiers must have at least one key-value pair).";
```
mikechu-optimizely, line 73: I'm open to adding new (more descriptive) messages as long as @jaeopt 's cool with it.

OptimizelySDK/Odp/OdpEventManager.cs
```diff
@@ -380,6 +386,20 @@ public void UpdateSettings(OdpConfig odpConfig)
...
+        /// </summary>
+        /// <param name="Identifiers">Identifiers to be analyzed</param>
+        /// <returns>True if identifiers dictionary is not null or empty otherwise False</returns>
+        private bool IsIdentifiersValid(Dictionary<string, string> Identifiers) 
+        {
+            if (Identifiers == null || Identifiers.Count == 0)
+            {
+                return false;
+            }
+            return true;
```
mikechu-optimizely, line 400: Would this work?

```csharp
return Identifiers?.Count > 0;
```

mikechu-optimizely (changes requested):
I have some thoughts below.

OptimizelySDK/Odp/Constants.cs
mnoman09: event data is separate and identifier is is not part of event data.

mikechu-optimizely:
@mnoman09, This is ready to merge whenever you'd like.

PR #343, opened 2023-04-07
[FSSDK-9039] Fix e2e ODP event response

mikechu-optimizely:
## Summary
As a bug bash engineer, I would like ODP events being sent from the C# SDK to not return an HTTP Status 400.

## Test plan
- Existing unit and integration test should pass
- Bug bash engineer's test app should pass end-to-end testing

## Issues
- FSSDK-9039

### DIFF
--- a/OptimizelySDK/Odp/OdpEventApiManager.cs
+++ b/OptimizelySDK/Odp/OdpEventApiManager.cs
@@ -92,7 +92,10 @@ namespace OptimizelySDK.Odp
             var endpoint = $"{apiHost}{Constants.ODP_EVENTS_API_ENDPOINT_PATH}";
             var data = JsonConvert.SerializeObject(events, settings: new JsonSerializerSettings
             {
-                ContractResolver = new CamelCasePropertyNamesContractResolver(),
+                ContractResolver = new DefaultContractResolver
+                {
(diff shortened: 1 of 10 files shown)


mikechu-optimizely (commented):
Developer PR review completed.

OptimizelySDK/Odp/OdpEventApiManager.cs
```diff
@@ -89,7 +90,10 @@ public bool SendEvents(string apiKey, string apiHost, List<OdpEvent> events)
             }
 
             var endpoint = $"{apiHost}{Constants.ODP_EVENTS_API_ENDPOINT_PATH}";
-            var data = JsonConvert.SerializeObject(events);
+            var data = JsonConvert.SerializeObject(events, settings: new JsonSerializerSettings
+            {
+                ContractResolver = new CamelCasePropertyNamesContractResolver(),
```
andrewleap-optimizely, line 95: Looking through the ODP docs, it looks like they are using snake case for their field names. Maybe we should consider future proofing by utilizing a snake case strategy instead?

andrewleap-optimizely (approved):
LGTM

pulak-opti (approved):
LGTM!

PR #344, opened 2023-04-11
[FSSDK-9093] Add return Task handle for async FetchQualifiedSegments

mikechu-optimizely:
## Summary
Async FetchQualifiedSegments should return a thread handle and is guaranteed to run to completion when `.Wait()` is used by the consumer.

## Test plan
- Existing unit and FSC tests should pass

## Issues
- FSSDK-9093


andrewleap-optimizely (approved):
LGTM

PR #347, opened 2023-04-17
[FSSDK-9090] Default instantiation of OdpManager for Optimizely constructors

mikechu-optimizely:
## Summary
- Fix OdpManager not having a default instantiation in `Optimizely` client

`OptimizelyFactory` provides a default which is desirable.  The `Optimizely` constructors should follow this.

## Test plan

- Add covering and passing unit test for `Optimizely`
- FSC integration tests should continue to pass

## Issues
- FSSDK-9090

### DIFF
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -51,6 +51,7 @@ namespace OptimizelySDK
     public class Optimizely : IOptimizely, IDisposable
     {
         private Bucketer Bucketer;
+
         [Obsolete]
         private EventBuilder EventBuilder;
 
@@ -222,8 +223,8 @@ namespace OptimizelySDK
 
             if (ProjectConfigManager.SdkKey != null)
             {
-                NotificationCenterRegistry.GetNotificationCenter(configManager.SdkKey, logger)?.
-                    AddNotification(NotificationCenter.NotificationType.OptimizelyConfigUpdate,
+                NotificationCenterRegistry.GetNotificationCenter(configManager.SdkKey, logger)
+                    ?.AddNotification(NotificationCenter.NotificationType.OptimizelyConfigUpdate,
                         () =>
                         {
                             projectConfig = ProjectConfigManager.CachedProjectConfig;
@@ -270,7 +271,7 @@ namespace OptimizelySDK
             OdpManager = odpManager ?? new OdpManager.Builder()
                 .WithErrorHandler(errorHandler)
                 .WithLogger(logger)
-                .Build();       
+                .Build();
 #endif
         }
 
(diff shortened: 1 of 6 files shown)


OptimizelySDK/Optimizely.cs
```diff
@@ -438,8 +442,8 @@ private bool ValidateInputs(string datafile, bool skipJsonValidation)
             userAttributes = userAttributes ?? new UserAttributes();
 
             var userContext = CreateUserContextCopy(userId, userAttributes);
-            var variation = DecisionService.GetVariation(experiment, userContext, config)?.
-                ResultObject;
+            var variation = DecisionService.GetVariation(experiment, userContext, config)
+                ?.ResultObject;
```
andrewleap-optimizely, line 446: I like the use of the leading period in method chaining- what do you think about leaving the null conditional operator on the previous line, grouping it with its relevant method?

andrewleap-optimizely (approved):
LGTM (with small nit)

OptimizelySDK/Optimizely.cs
andrewleap-optimizely: 🤷‍♂️ you win some, you lose some

mnoman09 (changes requested):
Overall changes lgtm. But how are we going to disable the odp? Previously we assumed that if user is not passing odpManager then it will be considered as disabled. In my opinion if someone is using OptimizelyFactory to initialize Optimizely then odpManager should be enabled by default otherwise do not initialize it and consider it disabled. 
@jaeopt What do you suggest?

jaeopt:
> I have a question. How are we going to disable the odp? Previously we assumed that if user is not passing odpManager then it will be considered as disabled. In my opinion if someone is using OptimizelyFactory to initialize Optimizely then odpManager should be enabled by default otherwise do not initialize it and consider it disabled. @jaeopt What do you suggest?

ODP will be enabled unless they

mikechu-optimizely:
@mnoman09 I updated the https://github.com/optimizely/csharp-testapp/pull/83

mikechu-optimizely:
Is it ok to have this PR open @jaeopt given my change to the test app to support FSC?

jaeopt:
> Is it ok to have this PR open @jaeopt given my change to the test app to support FSC?

@mikechu-optimizely oops, did I close the PR? I was wondering my comment has been broken and not realized I closed it by mistake. Sorry :)

mikechu-optimizely:
Sweet. Passing now.  I just hope I satisifed @mnoman09 's question to get a LGTM

mikechu-optimizely:
Hey @mnoman09, I think you might be out, but I'm going to go ahead and merge this change.  DM me if you wanna talk more about it.  I'm happy to open more PRs to get it right.

PR #348, opened 2023-04-24
[FSSDK-9128] fix: Handle ODP INVALID_IDENTIFIER_EXCEPTION code

mikechu-optimizely:
## Summary

Support a new format for ODP `InvalidIdentifierException` (when SDK fetches segments with a non-registered userId).

```
       "extensions": {
            "code": "INVALID_IDENTIFIER_EXCEPTION",
            "classification": "DataFetchingException"
       }
```

## Test plan
- Added unit tests to cover the new response format.
- Existing unit tests should pass
- FSC should continue to pass

## Issues
- FSSDK-9128

### DIFF
--- a/OptimizelySDK.Tests/OdpTests/OdpSegmentApiManagerTest.cs
+++ b/OptimizelySDK.Tests/OdpTests/OdpSegmentApiManagerTest.cs
@@ -147,7 +147,7 @@ namespace OptimizelySDK.Tests.OdpTests
                 response.Errors[0].Extensions.Code
             );
         }
-        
+
         [Test]
         public void ShouldParseOnlyFirstErrorInvalidIdentifierExceptionResponse()
         {
@@ -204,9 +204,11 @@ namespace OptimizelySDK.Tests.OdpTests
 
 
             Assert.IsNull(segments);
-            _mockLogger.Verify(l=>l.Log(LogLevel.WARN, "Audience segments fetch failed (invalid identifier)"), Times.Once);
+            _mockLogger.Verify(
+                l => l.Log(LogLevel.WARN, "Audience segments fetch failed (invalid identifier)"),
+                Times.Once);
         }
-        
+
         [Test]
         public void ShouldParseOnlyFirstErrorThatOdpThrowsAtUsResponse()
         {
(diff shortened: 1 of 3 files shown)


jaeopt (changes requested):
I do not see a logic change to provide proper error message for INVALID_IDENTIFIER_EXCEPTION. Can you check again?

mikechu-optimizely:
> I do not see a logic change to provide proper error message for `INVALID_IDENTIFIER_EXCEPTION`. Can you check again?

I'm not handling any specific ODP exceptions specially. I'm simply concatenating all current and future messages returned from Zaius (via `.ToString`) into one `errors`: string in [OdpSegmentApiManager.cs](https://github.com/optimizely/csharp-sdk/blob/mike/invalid-identifier/OptimizelySDK/Odp/OdpSegmentApiManager.cs#L120) 

Should I check-for and return a specific new message when I receive the `code == INVALID_IDENTIFIER_EXCEPTION` or `classification == DataFetchingException`?

OptimizelySDK.Tests/OdpTests/OdpSegmentApiManagerTest.cs
```diff
@@ -110,38 +110,165 @@ public void ShouldHandleAttemptToDeserializeInvalidJsonResponse()
...
+                            {
+                                ""line"": 4,
+                                ""column"": 5
+                            }
+                        ],
+                        ""path"": [
+                            ""desks/wooden""
+                        ],
+                        ""extensions"": {
+                            ""code"": ""SECOND_IGNORED_UNPLANNED_EXCEPTION"",
```
jaeopt, line 245: This "code" is optional. Wondering if we have a test case without "code".

jaeopt (commented):
Looks good!
The error "code" is optional. Can you check if we have a test case without code?

OptimizelySDK.Tests/OdpTests/OdpSegmentApiManagerTest.cs
jaeopt: ah I see no code in line 229.

PR #362, opened 2023-07-25
[FSSDK-9533] Log error & reject on track event with null, empty or whitespace event key

mikechu-optimizely:
## Summary
- Prevent track events from being sent without a name/key 
- Update method docs to highlight that this param must not be whitespace, null, or empty string values

## Test plan
- Updated existing tests to check for invalid event keys
- All other existing tests should also pass

## Issues
- FSSDK-9533

### DIFF
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -338,7 +338,7 @@ namespace OptimizelySDK
         /// <summary>
         /// Sends conversion event to Optimizely.
         /// </summary>
-        /// <param name="eventKey">Event key representing the event which needs to be recorded</param>
+        /// <param name="eventKey">Event key representing the event (must not be null or empty)</param>
         /// <param name="userId">ID for user</param>
         /// <param name="userAttributes">Attributes of the user</param>
         /// <param name="eventTags">eventTags array Hash representing metadata associated with the event.</param>
--- a/OptimizelySDK/Optimizely.cs
+++ b/OptimizelySDK/Optimizely.cs
@@ -87,7 +87,7 @@ namespace OptimizelySDK
         {
             get
             {
-                // Example output: "2.1.0" .  Should be kept in synch with NuGet package version.
+                // Example output: "2.1.0". Should be kept in sync with NuGet package version.
 #if NET35 || NET40
                 var assembly = Assembly.GetExecutingAssembly();
 #else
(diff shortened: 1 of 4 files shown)


OptimizelySDK/Optimizely.cs
```diff
@@ -338,54 +339,60 @@ private bool ValidateInputs(string datafile, bool skipJsonValidation)
...
         /// <param name="userId">ID for user</param>
         /// <param name="userAttributes">Attributes of the user</param>
         /// <param name="eventTags">eventTags array Hash representing metadata associated with the event.</param>
         public void Track(string eventKey, string userId, UserAttributes userAttributes = null,
             EventTags eventTags = null
         )
         {
-            var config = ProjectConfigManager?.GetConfig();
-
-            if (config == null)
```
jaeopt, line 351: can we remove this config ready check?

OptimizelySDK/Optimizely.cs
```diff
@@ -338,54 +339,60 @@ private bool ValidateInputs(string datafile, bool skipJsonValidation)
...
+            };
 
             if (!ValidateStringInputs(inputValues))
             {
                 return;
             }
 
-            var eevent = config.GetEvent(eventKey);
-
-            if (eevent.Key == null)
```
jaeopt, line 367: I think these 2 lines will filter out null or empty eventKey cases.

jaeopt (commented):
Can we create a test case to reproduce the suspected failure cases first? The current solution looks fine and also validated with FSC tests as well.

OptimizelySDK/Optimizely.cs
mikechu-optimizely: I think I see what you're talking about: The eventKey must be in the datafile before it can be tracked anyway, regardless of whether it's not null, empty, or whitespace.

OptimizelySDK/Optimizely.cs
mikechu-optimizely: If we're going to call `.GetEvent(eventKey)` later (L365 below on this older version) to check if the provided `eventKey` is in the datafile, I think we'd probably still need this check, right?

OptimizelySDK/Optimizely.cs
```diff
@@ -338,54 +339,60 @@ private bool ValidateInputs(string datafile, bool skipJsonValidation)
         /// <summary>
         /// Sends conversion event to Optimizely.
         /// </summary>
-        /// <param name="eventKey">Event key representing the event which needs to be recorded</param>
+        /// <param name="eventKey">Event key representing the event (must not be null, empty, or whitespace)</param>
```
mikechu-optimizely, line 342: I guess this should read: "Event key representing the event in the datafile which must not be null, empty, or whitespace"

OptimizelySDK/Optimizely.cs
```diff
@@ -338,54 +339,60 @@ private bool ValidateInputs(string datafile, bool skipJsonValidation)
...
         /// <param name="userAttributes">Attributes of the user</param>
         /// <param name="eventTags">eventTags array Hash representing metadata associated with the event.</param>
         public void Track(string eventKey, string userId, UserAttributes userAttributes = null,
             EventTags eventTags = null
         )
         {
-            var config = ProjectConfigManager?.GetConfig();
-
-            if (config == null)
+            if (eventKey == null || Regex.IsMatch(eventKey, @"^\s*$"))
```
mikechu-optimizely, line 350: My reason for adding this earlier is because it's slightly cheaper to test the eventKey is a good string before digging into the ProjectConfig to see if the eventKey is present in the datafile since it's [required per the docs](https://docs.developers.optimizely.com/feature-experimentation/docs/track-event-csharp#:~:text=Description-,eventKey%0Arequired,-String).

PR #369, opened 2023-11-21
[FSSDK-9775] Log invalid user attributes

mikechu-optimizely:
## Summary

- Log discovered invalid user attributes to an optional logger when BuildAttributeList

## Test plan

- Existing tests should pass
- Add new test coverage

## Issues
- FSSDK-9775

### DIFF
--- a/OptimizelySDK/Event/EventFactory.cs
+++ b/OptimizelySDK/Event/EventFactory.cs
@@ -193,24 +193,33 @@ namespace OptimizelySDK.Event
 
             var attributesList = new List<VisitorAttribute>();
             var invalidUserAttributeKeys = new List<string>();
+            var notRegisteredAttributeKeys = new List<string>();
 
             if (userAttributes != null)
             {
                 foreach (var userAttribute in userAttributes)
                 {
-                    if (Validator.IsUserAttributeValid(userAttribute))
+                    var attributeId = config.GetAttributeId(userAttribute.Key);
+                    if (string.IsNullOrEmpty(attributeId))
                     {
-                        var attributeId = config.GetAttributeId(userAttribute.Key);
-                        if (!string.IsNullOrEmpty(attributeId))
-                        {
-                            attributesList.Add(new VisitorAttribute(attributeId, userAttribute.Key,
-                                CUSTOM_ATTRIBUTE_FEATURE_TYPE, userAttribute.Value));
-                        }
+                        notRegisteredAttributeKeys.Add(userAttribute.Key);
+                        continue;
                     }
-                    else
+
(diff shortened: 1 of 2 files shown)


OptimizelySDK/Event/EventFactory.cs
```diff
@@ -192,19 +192,32 @@ ProjectConfig config
...
-                foreach (var validUserAttribute in userAttributes.Where(attribute =>
-                             Validator.IsUserAttributeValid(attribute)))
+                foreach (var userAttribute in userAttributes)
                 {
-                    var attributeId = config.GetAttributeId(validUserAttribute.Key);
-                    if (!string.IsNullOrEmpty(attributeId))
+                    if (Validator.IsUserAttributeValid(userAttribute))
+                    {
+                        var attributeId = config.GetAttributeId(userAttribute.Key);
+                        if (!string.IsNullOrEmpty(attributeId))
```
jaeopt, line 204: What about switching 204 and 201, so we only log invalid type attributes for attribute ids defined in datafile.
We do not care about types of not-registered attributes.

jaeopt (changes requested):
A suggestion. We can discuss about the options.

OptimizelySDK/Event/EventFactory.cs
mikechu-optimizely: Let me know what you think @jaeopt

mikechu-optimizely:
Looks like I'm going to need to work on the C# test app to pass FSC

OptimizelySDK/Event/EventFactory.cs
```diff
@@ -192,18 +192,40 @@ ProjectConfig config
...
                     }
+
+                    attributesList.Add(new VisitorAttribute(attributeId, userAttribute.Key,
+                        CUSTOM_ATTRIBUTE_FEATURE_TYPE, userAttribute.Value));
+                }
+
+                if (notRegisteredAttributeKeys.Count > 0 && !(logger is null))
+                {
+                    logger.Log(LogLevel.WARN,
+                        $"User attributes: {string.Join(", ", notRegisteredAttributeKeys.ToArray())} are not supported by the datafile and will not be used.");
```
jaeopt, line 222: Not sure if we want to log all these here. I see some clients have 100+ attributes not all registered to us. Each event will dump a bunch of these messages. Let's discuss.

jaeopt (changes requested):
The changes look good! 
One clarifying question about if we want to log all non-registered attributes or not.

mikechu-optimizely:
This was a spike only Jira ticket.  Woops. 🤷

PR #374, opened 2024-10-04
[FSSDK-10265] fix: UPS `Lookup` & `Save` during batched Decide

mikechu-optimizely:
## Summary

- User Profile Service should only `Lookup` and `Save` once when using `DecideAll` and `DecideForKeys`.

## Test plan

- Existing and new unit tests should pass
- FSC should pass

## Issues
- FSSDK-10265


OptimizelySDK/Bucketing/DecisionService.cs
```diff
@@ -43,6 +43,22 @@ public class DecisionService
...
+        private UserProfile _userProfile;
+
+        private bool _decisionBatchInProgress = false;
+
+        public bool DecisionBatchInProgress
+        {
+            get => _decisionBatchInProgress;
+            set
+            {
+                _decisionBatchInProgress = value;
```
mikechu-optimizely, line 55: Quick note for future Mike: You probably need to check that we're going from `true` to `false`.

OptimizelySDK/Bucketing/DecisionService.cs
```diff
@@ -467,29 +490,57 @@ UserProfile userProfile
...
+
+            if (!_decisionBatchInProgress)
+            {
+                SaveToUserProfileService(experiment, variation);
+            }
+        }
 
+        private void SaveToUserProfileService(Experiment experiment = null,
+            Variation variation = null
+        )
```
jaeopt, line 519: What about making experiment and variation NonNull and drop ?
I see this will be called only when a new decision is made.

OptimizelySDK/Bucketing/DecisionService.cs
```diff
@@ -140,35 +156,42 @@ OptimizelyDecideOption[] options
...
-            UserProfile userProfile = null;
+
             if (!ignoreUPS && UserProfileService != null)
             {
                 try
                 {
-                    var userProfileMap = UserProfileService.Lookup(user.GetUserId());
-                    if (userProfileMap != null &&
-                        UserProfileUtil.IsValidUserProfileMap(userProfileMap))
+                    if (_userProfile == null)
```
jaeopt, line 164: For legacy, we need to load fresh _userProfile on every getVariation() call (even if _userProfile not null)

OptimizelySDK/Optimizely.cs
```diff
@@ -1066,6 +1067,8 @@ OptimizelyDecideOption[] options
                 }
             }
 
+            DecisionService.DecisionBatchInProgress = false;
```
jaeopt, line 1070: I see this will fire the saveUPS. Wondering if we have the similar fire solution for legacy APIs (activate, getVarition)

OptimizelySDK.Tests/OptimizelyUserContextTest.cs
```diff
@@ -409,9 +424,55 @@ public void DecideWhenConfigIsNull()
             Assert.IsTrue(TestData.CompareObjects(decision, decisionExpected));
         }
 
-        #endregion decide
+        [Test]
+        public void DecideWithUspShouldOnlyLookupSaveOnce()
```
jaeopt, line 428:
```suggestion
        public void DecideWithUpsShouldOnlyLookupSaveOnce()
```

OptimizelySDK.Tests/OptimizelyUserContextTest.cs
```diff
@@ -443,6 +504,32 @@ public void DecideForKeysWithOneFlag()
...
+                    "We were unable to get a user profile map from the UserProfileService."),
+                Times.Never);
+            LoggerMock.Verify(
+                l => l.Log(LogLevel.ERROR, "The UserProfileService returned an invalid map."),
+                Times.Never);
+            userProfileServiceMock.Verify(l => l.Lookup(UserID), Times.Once);
+            userProfileServiceMock.Verify(l => l.Save(It.IsAny<Dictionary<string, object>>()),
+                Times.Once);
+        }
+
```
jaeopt, line 532: These tests look good. Wondering if we have tests covering the contents of the final UPS save call, combining all decisions in the session.

jaeopt: Can we also have the similar coverage (lookup once - save once, the ups save contents) with legacy APIs (activate, getVariation) for regression?

jaeopt (changes requested):
It looks good! 
I have a couple of suggestions for legacy API support and testing.

OptimizelySDK/Optimizely.cs
jaeopt: I see this comment not pushed properly:

  We probably need one more test -
  decide(A)
  decide(B)
  2nd session picks up the UPS as expected

OptimizelySDK.Tests/OptimizelyUserContextTest.cs
mikechu-optimizely: Added coverage of the final UPS `Save` as requested.  

I'm researching the `Activate` and `GetVariation` tests where UPS is used.

OptimizelySDK/Optimizely.cs
mikechu-optimizely: > I see this comment not pushed properly:
> 
> We probably need one more test - decide(A) decide(B) 2nd session picks up the UPS as expected

Added.

PR #409, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify events with single identifier

jaeopt:
## Summary

Block ODP identify events when only a single valid identifier is provided. Identify events exist to link multiple user identifiers together; sending them with a single identifier provides no value and wastes network calls.

**Jira:** [FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

### `OptimizelySDK/Odp/OdpEventManager.cs`
- Added filtering of null/empty identifier values using LINQ in `IdentifyUser`
- Added guard: requires 2+ valid identifiers to dispatch identify event
- Added debug log message when skipping: "ODP identify event is not dispatched (only one identifier provided)."

### Tests (`OdpEventManagerTests.cs`)
- Updated `ShouldLogWhenOdpNotIntegratedAndIdentifyUserCalled` to expect 2 ODP_NOT_INTEGRATED logs (instead of 3) since IdentifyUser now returns before reaching SendEvent
- Replaced `ShouldPrepareCorrectPayloadForIdentifyUser` with `ShouldNotSendIdentifyEventWithSingleIdentifier` to verify single-identifier identify events are blocked

## Notes
- C# SDK is server-side only (no VUID), so `IdentifyUser` always has a single `fs_user_id` identifier and identify events will be correctly skipped
- The public `IdentifyUser(string userId)` API is unchanged for backward compatibility
- The count check matches behavior across all SDKs per the spec

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/OptimizelySDK/Odp/OdpEventManager.cs
+++ b/OptimizelySDK/Odp/OdpEventManager.cs
@@ -383,6 +383,8 @@ namespace OptimizelySDK.Odp
                 .Where(kvp => !string.IsNullOrEmpty(kvp.Value))
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
 
+            // Identify requires 2+ identifiers to link (e.g., vuid + fs_user_id).
+            // A single identifier has no cross-reference value and generates unnecessary traffic.
             if (validIdentifiers.Count < 2)
             {
                 _logger.Log(LogLevel.DEBUG,
--- a/OptimizelySDK/Odp/OdpEventManager.cs
+++ b/OptimizelySDK/Odp/OdpEventManager.cs
@@ -378,7 +378,19 @@ namespace OptimizelySDK.Odp
                 { Constants.FS_USER_ID, userId },
             };
 
-            var odpEvent = new OdpEvent(Constants.ODP_EVENT_TYPE, "identified", identifiers);
+            // Filter out null/empty identifier values and require 2+ valid identifiers
+            var validIdentifiers = identifiers
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value))
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+
+            if (validIdentifiers.Count < 2)
+            {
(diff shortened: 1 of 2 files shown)


OptimizelySDK/Odp/OdpEventManager.cs
```diff
@@ -378,7 +378,21 @@ public void IdentifyUser(string userId)
...
 
-            var odpEvent = new OdpEvent(Constants.ODP_EVENT_TYPE, "identified", identifiers);
+            // Filter out null/empty identifier values and require 2+ valid identifiers
+            var validIdentifiers = identifiers
+                .Where(kvp => !string.IsNullOrEmpty(kvp.Value))
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+
+            // Identify requires 2+ identifiers to link (e.g., vuid + fs_user_id).
+            // A single identifier has no cross-reference value and generates unnecessary traffic.
+            if (validIdentifiers.Count < 2)
```
junaed-optimizely, line 388: @jaeopt, this condition will always evaluate to true in the current C# SDK. `IdentifyUser(string userId)` constructs a dictionary with exactly one entry (`fs_user_id`), so after filtering, `validIdentifiers.Count` is either 1 or 0. It can never reach 2.

This effectively makes `IdentifyUser` a permanent no-op. Is this intentional?

junaed-optimizely (approved):

OptimizelySDK/Odp/OdpEventManager.cs
jaeopt: @junaed-optimizely  yes a good catch - it's true for all server-sdks. We had an option to clean up all but decide to keep this way since we have a chance to change this behavior or add other identities.

PR #417, opened 2026-07-06
Junaed/fssdk 12756 remove authenticode step

junaed-optimizely:
## Summary

Removes the Authenticode signing step from the release pipeline.

Manual Authenticode signing of individual DLLs is no longer a standard practice for libraries distributed via NuGet. 

The .NET ecosystem has shifted toward NuGet's own package signing (author and repository signatures) as the primary trust and integrity mechanism, making a separate DLL-level Authenticode step unnecessary overhead.

The signing job in our pipeline was also never operational. It contained placeholder paths and an explicit halt, so this removal reflects both an industry shift and the cleanup of dead infrastructure.

## Changes
- Removed the sign job from csharp_release.yml
- Updated pack to depend directly on the build jobs and consume DLL artifacts
- Added nuget/setup-nuget@v2 setup step

## Issues
- FSSDK-12756

### DIFF
--- a/.github/workflows/csharp_release.yml
+++ b/.github/workflows/csharp_release.yml
@@ -178,7 +178,7 @@ jobs:
         run: |
           pushd ./dlls
           # Move all dlls to the root directory
-          find . -type f -name "*.dll" -exec mv {} .
+          find . -type f -name "*.dll" | xargs -I{} mv {} .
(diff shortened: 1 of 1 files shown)


muzahidul-opti (approved):
LGTM

